PR #54, opened 2015-01-30
 Fixes #53 - js error if _spoor is not set in config.json

### LINKED ISSUE #53: Spoor generates an error that _isEnabled is not defined
Old courses will not have an isEnabled property.

Some defensive programming is required on this line:
https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/js/adapt-contrib-spoor.js#L25

dennis-learningpool:


olivermartinfoster:
+1 cool

olivermartinfoster:
(we should probably make sure all of the extensions fall-over nicely when they configurations are missing)

brian-learningpool:
+1

taylortom:
+1, and +1 for Ollie's suggestion

moloko:
would it not also be necessary to check for the existence of _isEnabled as well?

I thought that not existing was the problem @brian-learningpool  was highlighting here: https://github.com/adaptlearning/adapt-contrib-spoor/issues/53

dennis-learningpool:
Hi Matt,

The issue brian highlights is that if `this.data` is null, then attempting to access `this.data._isEnabled` will result in a Javascript error. As long as the existence of `this.data` is confirmed first, there will be no error in accessing the `_isEnabled` property, whether it is set or not.

moloko:
OK I see now

+1 in that case

PR #57, opened 2015-03-19
Lukasz Grela - ScormWrapper settings externalised

LukaszGrela:
Externalisation of the `ScormWrapper`'s settings, those settings are in the `config.json` in `_spoor` object under `_scormWrapper` object property.

	"_spoor": {
        "_scormWrapper": {
            "showDebug":true,
            "commitOnStatusChange":true,
            "disableInteractionTracking":false,
            "timedCommitFrequency":10,
            "maxCommitRetries":5,
            "commitRetryDelay":2000
        }
	}

### DIFF
--- a/example.json
+++ b/example.json
@@ -11,6 +11,15 @@
 			"_comment": "Your options here are 'completed', 'passed', 'failed', and 'incomplete'",
 			"_onTrackingCriteriaMet": "completed",
 			"_onAssessmentFailure": "incomplete"
-		}
+		},
+	        "_scormWrapper": {
+	            "setCompletedWhenFailed":true,
+	            "commitOnStatusChange":true,
+	            "disableInteractionTracking":false,
+	            "timedCommitFrequency":10,
+	            "maxCommitRetries":5,
+	            "commitRetryDelay":2000,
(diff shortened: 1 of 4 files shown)


moloko:
good idea!

however, `commitRetryPending` is effectively a private variable of scormWrapper.js and shouldn't be exposed in this way

I would also suggesting changing `_scormWrapper` in the json to something a little more user-friendly and descriptive like `_advancedSettings`

The property `setCompletedWhenFailed` really belongs in the _tracking section of the JSON as it's related specifically to tracking - although to be honest since this relates only to SCORM 2004 (which we're not officially supporting yet) it's maybe best to leave this out.

I suspect the authoring tool team would also like the properties.schema to be updated with all of this new data.

Finally, if you have time, it would also be great to have a setting to force the debug window to open automatically (see `scormWrapper.showDebugWindow()`) as currently you can only do this by adding a variable __debug to index.html

LukaszGrela:
I presume the validators for a number values in schema should be added but I dont know what options are possible

moloko:
as discussed, this needs amending so that it works with the latest version of the develop branch. I will try and find some time to do this.

PR #58, opened 2015-03-20
Lukasz Grela - remember cancel response

LukaszGrela:
This is "which-part-of-cancel-you-dont-understand" fix :) So when users choose 'cancel' it is remembered and user is not bothered any more.

### DIFF
--- a/js/scormWrapper.js
+++ b/js/scormWrapper.js
@@ -56,6 +56,9 @@ define (function(require) {
 
 		this.registeredViews = [];
 		
+		//flag that is set to true when user has clicked 'cancel' in the "confirm" dialogue
+        	this.userRejectedDebugWindow = false;
+        
 		if (window.__debug)
 			this.showDebugWindow();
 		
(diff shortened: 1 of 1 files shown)


hrajotia:
+1

moloko:
I'm not sure this is a good idea as it's basically allowing the user to suppress potentially serious tracking errors.

If you find the debug window annoying during development, you can either temporarily disable tracking via config.json - or just press OK to open the debug window. It'll just sit in the background logging messages.

LukaszGrela:
The message is not obvious, and usually this is annoying on mobile when popups are disabled, then this popup appears no matter what you choose.

taylortom:
I tend to agree with @moloko; I think we need to keep some kind of notification to warn the user something's amiss, I don't feel that ignoring all further notifications is the best solution.

A compromise could be to check the course is connected to an LMS when it's first opened, then give a single error at that point. If it connects successfully and errors are thrown later, we need to be notified of these as and when they occur.

brian-learningpool:
I agree with @taylortom and @moloko.

PR #60, opened 2015-04-02
Adds a defer to setting the components to complete

### LINKED ISSUE #59: Spoor has an issue where the serializer tries to set completion on blocks before the global events are attached
During load of a course in Moodle or Totara a course can return with completed components but the blocks and articles are not set to complete due to the events on the models not being setup in time.

Time for a trusty _.defer

darylhedley:
fixes #59

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "adapt-contrib-spoor",
-  "version": "1.1.8",
+  "version": "1.1.9",
   "homepage": "https://github.com/adaptlearning/adapt-contrib-spoor",
   "authors": [
     "Kevin Corry <kevinc@learningpool.com>",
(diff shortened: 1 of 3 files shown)


brian-learningpool:
+1

moloko:
+1 although I'm surprised we've not hit this issue given that we test all our courses on Moodle! Can you also PR to develop once this is done?

oliverfoster:
+1 this is an issue with the app:dataReady event and the old 'order in which events execute' problem. both adaptModel's.setupModels and spoor's.dataReady both listen to the app:dataReady event. perhaps we need an adapt:dataReady, or an app:modelsReady inbetween app:dataReady and adapt:initialize?

brian-learningpool:
Ollie, that's an idea.  Perhaps there should be separate events for the core bootstrapping before the framework signals events which plugins can tap into such as app:dataReady?

oliverfoster:
It'd certainly be nice to have some more space in that arena. I quite like the 'pre' and 'post' event styling, although it might be a bit ambiguous for this case as nothing is really rendering but the models.

PR #72, opened 2015-07-13
ABU-940: Properly disable offlineStorage implementation

oliverfoster:
* when spoor_isEnabled: false, stop offlineStorage API calling SCORM functions

### DIFF
--- a/js/adapt-offlineStorage-scorm.js
+++ b/js/adapt-offlineStorage-scorm.js
@@ -20,9 +20,7 @@ define([
 		get: function(name) {
 			if (name === undefined) {
 				//If not connected return just temporary store.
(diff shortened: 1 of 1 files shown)


moloko:
+1

moloko:
that said, would this:
{{{
useTemporaryStore: function() {
	var cfg = Adapt.config.get('_spoor');

	if (!scorm.lmsConnected || !cfg || cfg._isEnabled === false) return true;
	return false;
}
}}}
Not be a slightly tidier way of doing it?

oliverfoster:
Yea, nice spot, I'll change it in the morning.

hrajotia:
+1

moloko:
+1

zenduo:
+1

PR #75, opened 2015-07-17
Decimal array fix

oliverfoster:
* Update to always choose to convert to a decimal array when the numeric array being saved contains a decimal value.

When a text input with 2 input boxes has answers of  1 correct and 1 incorrect it produces the array > [0,-1]

Arrays must have a common type in order to be saved in this format. Therefore a numeric array must be normalized before it is saved.

As decimal is the only numeric type (from half, byte, short, long and decimal) to accommodate negative numbers:

[0, -1] would be a decimal array
[half, decimal] > [decimal, decimal]

have forced this behaviour


tomgreenfield:
+1

dancgray:
+1

hrajotia:
+1

PR #77, opened 2015-07-22
fix incorrect use of 'this' & correct attempt to call method when should be setting a property

moloko:
fixes https://adaptlearning.atlassian.net/browse/ABU-985


dennis-learningpool:
+1

lc-thomasberger:
+1

brian-learningpool:
+1

PR #79, opened 2015-07-29
Fix: initialize at config load not course load

oliverfoster:
* Allows for language to be saved / restored before the course data has loaded
* Allows for multiple language folders and pre-load language selection such as > https://github.com/cgkineo/adapt-languageSelection/blob/master/js/langsel.js


moloko:
+1

belal-almassri-spongeuk:
+1 Thank you very much Ollie, now I have a fully working multiple language select course.

taylortom:
+1

PR #82, opened 2015-08-06
Fix: made bit count a factor of 4

oliverfoster:
Partly fixes issue with textinput and three text boxes with generic answers


moloko:
+1

dennis-learningpool:
+1

belal-almassri-spongeuk:
+1

Issue #83, opened 2015-08-07
questions serializer bug


oliverfoster:
* doesn't restore data to same set of components, uses all components in block to restore to rather than just the question components this pushes off the block locations

belal-almassri-spongeuk:
This works fine now.

PR #84, opened 2015-08-07
fix: use only questions in capture and release

oliverfoster:
* doesn't restore data to same set of components, uses all components in block to restore to rather than just the question components this pushes off the block locations

#83


moloko:
+1

belal-almassri-spongeuk:
+1

zenduo:
+1

PR #96, opened 2015-09-04
Schema defaults and version bump.

dancgray:


oliverfoster:
+1

oliverfoster:
@moloko this'll be one you'll want to check

properties.schema
```diff
@@ -114,6 +131,7 @@
                     "_commitOnStatusChange": {
                       "type":"boolean",
                       "required":false,
+                      "default": "false",
```
moloko: This should be `true`

properties.schema
```diff
@@ -32,6 +33,7 @@
                     "_requireCourseCompleted": {
                       "type":"boolean",
                       "required":false,
+                      "default": "false",
```
moloko: I think one of the completion criteria needs to be on by default, otherwise using the default settings will mean the course never gets set to completed. My suggestion would be for `_requireCourseCompleted` to be `true` by default

moloko:
@dancgray, while you're at it, could you change the help for `_scormVersion` to:
> Which SCORM version this course is compatible with. Note that currently only 1.2 is officially supported.

brian-learningpool:
This being the case, should we just remove option for 2004 until it's actually supported?

moloko:
I'm fine with leaving it in. 'not supported' essentially means 'it almost certainly works but we don't test it' and 'you'll have to add the relevant packaging files manually'

brian-learningpool:
If that's the case, I'd actually still prefer to take it out of the schema.  The properties.schema is specifically read by the authoring tool and I think a dependency on adding the relevant packaging files manually is counter-intuitive to the goals of the authoring tool.

moloko:
OK fine by me

dancgray:
@moloko , @brian-learningpool. Have changed the defaults as per recommendations.

oliverfoster:
+1

moloko:
still need to remove _scormVersion from the schema as per @brian-learningpool 's request

oliverfoster:
the 2004 entry has been removed, should that whole section go missing?

dancgray:
If it's missing from the properties.schema it will be missing from the authoring tool output JSON.  If we don't need "_scormVersion": "1.2", in the output the whole attribute can be removed otherwise I'll have a look to see if it can be hidden as there is not much point having only one option in the builder form.

moloko:
you don't need to have `_scormVersion` in the output. that whole `_advancedSettings` section is optional

dancgray:
_scormVersion removed, looks like a default is set in the code anyway https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/js/adapt-contrib-spoor.js#L51

moloko:
+1

hrajotia:
+1

brian-learningpool:
+1

PR #100, opened 2015-09-07
Fixes #98, adds defensive programming

### LINKED ISSUE #98: Defensive programming missing from adapt-stateful-session.js
The code in the following line doesn't check that the spoor plugin is actually enabled on a course.  There should possibly be a `.has()` call here.
https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/js/adapt-stateful-session.js#L24

This manifests itself in errors by default in the authoring tool

brian-learningpool:
If the spoor extension is not enabled, the code should not throw errors because the relevant configruation does not exist.  I've added some checks for this.

### DIFF
--- a/js/adapt-contrib-spoor.js
+++ b/js/adapt-contrib-spoor.js
@@ -33,8 +33,12 @@ define([
 		},
 
 		checkConfig: function() {
-			this._config = Adapt.config.get('_spoor');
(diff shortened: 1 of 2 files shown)


hrajotia:
+1

moloko:
+1

taylortom:
+1

PR #108, opened 2015-10-12
adapt-contrib-spoor.js: ensure that `setVersion` is called even when 

moloko:
`_advancedSettings` are not set (fixes ([https://adaptlearning.atlassian.net/browse/ABU-1076](ABU-1076))

wrapper.js: refactor `setVersion` so that it uses the pre-existing `isSCORM2004` function


brian-learningpool:
+1

oliverfoster:
+1

LukaszGrela:
+1

PR #109, opened 2015-10-22
The "silent" property

LukaszGrela:
Add silent property that will prevent (when true) appearance of any error popups along with debug window without disabling spoor. Useful when the same build must be used in LMS and non-LMS environment.

### DIFF
--- a/example.json
+++ b/example.json
@@ -2,6 +2,7 @@
 {
 	"_spoor": {
 		"_isEnabled": true,
+        	"_silent": true,
 		"_tracking": {
 			"_requireCourseCompleted": true,
 			"_requireAssessmentPassed": false,
--- a/example.json
+++ b/example.json
@@ -2,7 +2,6 @@
 {
 	"_spoor": {
 		"_isEnabled": true,
-        	"_silent": true,
 		"_tracking": {
 			"_requireCourseCompleted": true,
 			"_requireAssessmentPassed": false,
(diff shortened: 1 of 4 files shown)


moloko:
please see my comments on #110

LukaszGrela:
The idea is to intentionally make it silent, if one care about showing to the user info about the LMS issues simply not use this property.

example.json
```diff
@@ -21,7 +21,8 @@
 			"_commitOnStatusChange": false,
 			"_timedCommitFrequency": 10,
 			"_maxCommitRetries": 5,
-			"_commitRetryDelay": 2000
+			"_commitRetryDelay": 2000,
+        		"_suppressErrors": true
```
moloko: would you mind just setting it to `false` here?

properties.schema
```diff
@@ -150,6 +150,15 @@
...
                       "help": "How much of a delay (in milliseconds) to leave between commit retries."
+                    },
+                    "_suppressErrors": {
+                      "type":"boolean",
+                      "required":false,
+                      "default": "false",
+                      "title":"Supress LMS errors",
+                      "inputType": {"type": "Boolean", "options": [true, false]},
+                      "validators": [],
+                      "help": "Whether spoor plugin should show any errors to the user."
```
moloko: Could you change this to 'Set this to 'true' to stop this plugin from displaying error messages when tracking problems occur.'

moloko:
+1
lovely stuff, thanks

oliverfoster:
+1

taylortom:
+1

PR #110, opened 2015-10-22
The "silent" property - master

LukaszGrela:
Add silent property that will prevent (when true) appearance of any error popups along with debug window without disabling spoor. Useful when the same build must be used in LMS and non-LMS environment.

### DIFF
--- a/example.json
+++ b/example.json
@@ -2,6 +2,7 @@
 {
 	"_spoor": {
 		"_isEnabled": true,
+        	"_silent": true,
 		"_tracking": {
 			"_requireCourseCompleted": true,
 			"_requireAssessmentPassed": false,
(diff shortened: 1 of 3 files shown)


moloko:
This setting seems a bit risky to me - what if something does go wrong? The user won't be notified and will carry on blithely thinking that what they're doing is being tracked...

That said, if we make the risks of using this setting clear in the README then I guess its OK to add.

However:
* this needs to go into develop, not master
* I'd prefer the property to be called `_suppressErrors`, this is more descriptive
* it should be in the `_advancedSettings` section
* the properties.schema needs updating to include it
* I don't think it should be used in line 61 of wrapper.js - if I switch debugging on by whatever method, I still want to see the SCORM debug window regardless. Although using `window.__debug` is now somewhat redundant since the introduction of `_showDebugWindow`, I think we should leave it alone for the sake of backwards compatibility

moloko:
The README will need updating as well but I'm happy to do that

moloko:
sorry, I've just seen you've done a PR against develop as well. You can ditch this one as I'll be merging develop into master across the board as soon as I have some time available!

PR #120, opened 2016-02-17
issue-957: isoptional and component completion restore

oliverfoster:
Allow spoor to restore component completion if block is already complete. isOptional fix.

### DIFF
--- a/js/serializers/default.js
+++ b/js/serializers/default.js
@@ -71,7 +71,7 @@ define([
         },
 
         markBlockAsComplete: function(block) {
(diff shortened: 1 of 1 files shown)


brian-learningpool:
+1

dancgray:
+1

moloko:
+1 
also gave it a test with a block containing:
* two components, one of which was optional
* two components, both of which were optional
* one component, which was optional
Seems fine in all those situations.

I really need to go and learn some more about writing unit tests because this exact situation screams 'should have a unit test for all this'!

PR #121, opened 2016-02-18
amend adapt-stateful-session.js so that it won't attempt to do interaction tracking

moloko:
if the question component hasn't been setup for it (see https://github.com/adaptlearning/adapt_framework/issues/955)


oliverfoster:
+1

dancgray:
+1

lc-thomasberger:
+1

PR #124, opened 2016-03-15
updates to allow for better validation

moloko:
see https://github.com/adaptlearning/adapt_framework/issues/997

remove the 'encoding' meta tag as it is not valid
remove 'language' attribute from script tags, these are not needed
remove 'role' attribute from button tag, it is not needed


oliverfoster:
+1 tabbing is off

moloko:
should be fixed now

hrajotia:
+1

brian-learningpool:
+1

PR #125, opened 2016-03-22
update to the latest pipwerks library

moloko:
pulled from https://github.com/pipwerks/scorm-api-wrapper


moloko:
not sure why it won't show the diff, only a few lines should have changed! will re-do this

PR #129, opened 2016-07-07
Add visibilitychange listener

tomgreenfield:
Uses the Page Visibility API to perform a SCORM commit whenever a course becomes hidden.

In my tests on Moodle, the `visibilitychange` event was firing successfully in the following:
* iOS 8
* iOS 9
* Android 6
* Safari 9
* Firefox 47

The event wasn't so consistent in:
* Chrome 51
* Edge 25
* IE 11

Although the original link mentions also implementing a `pagehide` listener, the `unload` handler was firing successfully in the cases where `pagehide` worked.

Original issue: https://github.com/adaptlearning/adapt_framework/issues/873.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,8 @@ The attributes listed below are used in *config.json* to configure **Spoor**, an
 
 >>**_commitRetryDelay** (number): Specifies the interval in milliseconds between commit retries. The default is `2000`.
 
+>>**_commitOnVisibilityChangeHidden** (boolean): Determines whether a "commit" call should be made whenever the course becomes hidden. The default is `true`.
+
 <div float align=right><a href="#top">Back to Top</a></div>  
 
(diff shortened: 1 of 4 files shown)


moloko:
nice one! will give this a good test some time over the next few days

moloko:
This is working really well for me in IE11, Edge, Firefox 48, Chrome, Safari 9 and iOS9. Very happy to go ahead with this - I just think it needs an option to allow developers to disable it because some Learning Management Systems have very specific requirements around when you should and shouldn't call `LMSCommit`.

I think it should be enabled by default - if someone wants to disable it then they should set  `_advancedSettings._commitOnVisibilityChangeHidden` to `false`.

moloko:
+1

brian-learningpool:
+1

oliverfoster:
+1

PR #132, opened 2016-08-18
multilang feature

lc-thomasberger:
make sure "lang" is always set in suspendDataStore
adds ability to reset session

### DIFF
--- a/js/adapt-offlineStorage-scorm.js
+++ b/js/adapt-offlineStorage-scorm.js
@@ -25,7 +25,8 @@ define([
 					location: scorm.getLessonLocation(),
 					score: scorm.getScore(),
 					status: scorm.getStatus(),
-					student: scorm.getStudentName()
+					student: scorm.getStudentName(),
+					learnerInfo: this.getLearnerInfo()
 				});
 
 				suspendDataRestored = true;
@@ -44,8 +45,10 @@ define([
 					return scorm.getScore();
 				case "status":
 					return scorm.getStatus();
-				case "student":
+				case "student":// for backwards-compatibility. learnerInfo is preferred now and will give you more information
 					return scorm.getStudentName();
+				case "learnerInfo":
+					return this.getLearnerInfo();
 				default:
 					return this.getCustomState(name);
 			}
@@ -81,7 +84,8 @@ define([
 				case "status":
 					return scorm.setStatus.apply(scorm, args);
 				case "student":
-					return false;
+				case "learnerInfo":
+					return false;// these properties are read-only
 				case "suspenddata":
 				default:
 					if (isObject) {
@@ -117,6 +121,33 @@ define([
 			
 			if (!scorm.lmsConnected || (cfg && cfg._isEnabled === false)) return true;
 			return false;
+		},
+
+		/**
+		 * Returns an object with the properties:
+		 * - id (cmi.core.student_id)
+		 * - name (cmi.core.student_name - which is usually in the format "Lastname, Firstname" - but sometimes doesn't have the space after the comma)
+		 * - firstname
+		 * - lastname
+		 */
+		getLearnerInfo: function() {
+			var name = scorm.getStudentName();
+			var firstname = "", lastname = "";
+			if (name && name !== 'undefined' && name.indexOf(",") > -1) {
+				//last name first, comma separated
+				var nameSplit = name.split(",");
+				lastname = $.trim(nameSplit[0]);
+				firstname = $.trim(nameSplit[1]);
+				name = firstname + " " + lastname;
+			} else {
+				console.log("SPOOR: LMS learner_name not in 'lastname, firstname' format");
+			}
+			return {
+				name: name,
+				lastname: lastname,
+				firstname: firstname,
+				id: scorm.getStudentId()
+			};
 		}
 		
 	});
--- a/js/adapt-offlineStorage-scorm.js
+++ b/js/adapt-offlineStorage-scorm.js
@@ -95,6 +95,7 @@ define([
 					}
 
 					var dataAsString = JSON.stringify(suspendDataStore);
+					// make sure the lang attribute used to store the selected language is always written into localstorage
 					return (suspendDataRestored || name.toLowerCase() === "lang") ? scorm.setSuspendData(dataAsString) : false;
 			}
 		},
--- a/js/adapt-offlineStorage-scorm.js
+++ b/js/adapt-offlineStorage-scorm.js
@@ -95,7 +95,7 @@ define([
 					}
 
 					var dataAsString = JSON.stringify(suspendDataStore);
-					return (suspendDataRestored) ? scorm.setSuspendData(dataAsString) : false;
+					return (suspendDataRestored || name.toLowerCase() === "lang") ? scorm.setSuspendData(dataAsString) : false;
 			}
 		},
 
--- a/js/adapt-offlineStorage-scorm.js
+++ b/js/adapt-offlineStorage-scorm.js
@@ -1,7 +1,7 @@
 define([
-	'coreJS/adapt',
+	'core/js/adapt',
 	'./scorm',
-	'coreJS/offlineStorage'
+	'core/js/offlineStorage'
 ], function(Adapt, scorm) {
 
 	//SCORM handler for Adapt.offlineStorage interface.
--- a/js/adapt-offlineStorage-scorm.js
+++ b/js/adapt-offlineStorage-scorm.js
@@ -95,7 +95,8 @@ define([
 					}
 
 					var dataAsString = JSON.stringify(suspendDataStore);
-					// make sure the lang attribute used to store the selected language is always written into localstorage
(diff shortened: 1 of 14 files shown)


js/adapt-stateful-session.js
```diff
@@ -79,6 +79,24 @@ define([
 			this.listenTo(Adapt.blocks, 'change:_isComplete', this.onBlockComplete);
 			this.listenTo(Adapt.course, 'change:_isComplete', this.onCompletion);
 			this.listenTo(Adapt, 'assessment:complete', this.onAssessmentComplete);
+			this.listenTo(Adapt, 'app:resetSession', this.onResetSession);
+		},
+
+		removeEventListeners: function () {
+			$(window).off('unload', this._onWindowUnload);
```
oliverfoster: Does this get reattached after reset?

js/adapt-stateful-session.js
```diff
@@ -127,6 +145,16 @@ define([
...
 		},
 
+		onResetSession: function () {
+			this.removeEventListeners();
+			this.setupEventListeners();
+			var sessionPairs = this.getSessionState();
+			
+			// hard status reset
+			Adapt.offlineStorage.set("status", "incomplete");
+			Adapt.offlineStorage.set(sessionPairs);
```
oliverfoster: Does this reset everything?

js/adapt-stateful-session.js
lc-thomasberger: onResetSession calls `this.setupEventListeners` this function calls `$(window).on('unload', this._onWindowUnload)`.

js/adapt-stateful-session.js
lc-thomasberger: when you reload the course json, it creates a new set of collections. This resets all the progress in the Model Data. That's why I also reset status to incomplete.

What I know by hard:
It will keep the Assessment object in offlinestorage. This can be used to restore session store for each language during multiple language switches.
If a Score is available, it will remain untio it is overwritten with new score.
Not sure about question interaction, will need to check

js/adapt-stateful-session.js
oliverfoster: Yep, sorry, it's just the binding you moved. Cool

oliverfoster:
seems good +1

js/adapt-stateful-session.js
```diff
@@ -79,6 +79,24 @@ define([
...
+			}
+
+			if(this._shouldRecordInteractions) {
+				this.stopListening(Adapt, 'questionView:recordInteraction', this.onQuestionRecordInteraction);
+			}
+
+			this.stopListening(Adapt.blocks, 'change:_isComplete', this.onBlockComplete);
+			this.stopListening(Adapt.course, 'change:_isComplete', this.onCompletion);
+			this.stopListening(Adapt, 'assessment:complete', this.onAssessmentComplete);
+			this.stopListening(Adapt, 'app:resetSession', this.onResetSession);
```
tomgreenfield: Could lines 87–99 be changed to simply:
```javascript
this.stopListening();
````
It looks like it's unnecessary to explicitly name all the callbacks here.

File: js/adapt-offlineStorage-scorm.js [hunk 4]
```diff
@@ -95,7 +95,7 @@ define([
 					}
 
 					var dataAsString = JSON.stringify(suspendDataStore);
-					return (suspendDataRestored) ? scorm.setSuspendData(dataAsString) : false;
+					return (suspendDataRestored || name.toLowerCase() === "lang") ? scorm.setSuspendData(dataAsString) : false;
```
moloko: This line could use an explanatory comment

js/adapt-stateful-session.js
```diff
@@ -127,6 +145,16 @@ define([
 			Adapt.offlineStorage.set("interaction", id, response, result, latency, responseType);
 		},
 
+		onResetSession: function () {
+			this.removeEventListeners();
+			this.setupEventListeners();
```
moloko: I don't understand 148 + 149 - why remove then immediately re-attach the event listeners?

js/adapt-stateful-session.js
lc-thomasberger: Good point, sounds sensible! Will change and give it a test.

js/adapt-stateful-session.js
oliverfoster: different collections

js/adapt-stateful-session.js
```diff
@@ -127,6 +145,16 @@ define([
...
+			var sessionPairs = this.getSessionState();
+			
+			// hard status reset
+			Adapt.offlineStorage.set("status", "incomplete");
```
moloko: I'm not 100% sure we should reset the lesson status here. At the very least I think it should probably be a config setting.
In fact, it's arguable that all this should be configurable - after all, if you did have a course where all languages have the same structure then technically there's no reason to reset.

js/adapt-stateful-session.js
lc-thomasberger: When you load a new language, it creates a new blocks, course, components ... Collection from the new JSON data. The Collection can be different for each language. So I thought it would be better to remove and reattach the eventListener.

js/adapt-stateful-session.js
oliverfoster: config.json: _spoor._reporting._onSessionReset: "incomplete" ?

js/adapt-stateful-session.js
```diff
@@ -79,6 +79,24 @@ define([
 			this.listenTo(Adapt.blocks, 'change:_isComplete', this.onBlockComplete);
 			this.listenTo(Adapt.course, 'change:_isComplete', this.onCompletion);
 			this.listenTo(Adapt, 'assessment:complete', this.onAssessmentComplete);
+			this.listenTo(Adapt, 'app:resetSession', this.onResetSession);
```
tomgreenfield: I had a hard time tracking down where this event gets triggered from. Eventually Ollie showed me it is in [Language Picker](https://github.com/adaptlearning/adapt-contrib-languagePicker/blob/master/js/languagePickerModel.js#L91).

Since we use Adapt's global events system, the prefix is normally used to signify its origin e.g. anything with `app:` I'd expect to be from core's `app.js`.

Suggest we rename this event to make it easier to track down in the future? `languagePicker:resetSession` seems the most logical.

oliverfoster: +1 but as ``spoor:resetSession`` this is an intention is to cause an action, not to notification of a state

js/adapt-stateful-session.js
lc-thomasberger: so one config to enable / disable reset and a value to use for resetting?

js/adapt-stateful-session.js
moloko: I'm with @tomgreenfield - the event originates from language picker and presumably is only triggered if the user chooses a different language.. shouldn't the event therefore be `languagePicker::languageChange`?

oliverfoster: Fair enough.

I'm thinking this is more of a "notify:*" situation https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/notify.js#L54-L68 where the behaviour is triggered by the endpoint event listener or spoor provides an API for other plugins to use.

I can see that the ``languagePicker:changeTrigger`` is much the same as "accessibility:toggle" https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/views/accessibilityView.js#L46

Maybe we should have left it as "app:sessionReset" which is a pretty reasonable space between the two where neither languagePicker, nor spoor take ownership?

oliverfoster:
+1

tomgreenfield:
Code looks good, +1.

Please remember to document `_resetStatusOnLanguageChange` and bump the version in README.

js/serializers/questions.js
```diff
@@ -1,5 +1,5 @@
 define([
-    'coreJS/adapt',
+    'core/js/adapt',
     './scormSuspendDataSerializer'
```
tomgreenfield: Realise this was present before your amends but is `'./scormSuspendDataSerializer'` actually used? It hasn't got an associated variable.

oliverfoster:
Yes, it has a global api. No reference needed.

js/serializers/questions.js
oliverfoster: It's fine to leave as is.

js/adapt-stateful-session.js
```diff
@@ -79,6 +77,17 @@ define([
 			this.listenTo(Adapt.blocks, 'change:_isComplete', this.onBlockComplete);
 			this.listenTo(Adapt.course, 'change:_isComplete', this.onCompletion);
 			this.listenTo(Adapt, 'assessment:complete', this.onAssessmentComplete);
+			this.listenTo(Adapt, 'languagePicker:languageChange', this.onLanguageChange);
+		},
```
brian-learningpool: The event name has changed to `app:languageChanged` in the `feature/ml` branch.

brian-learningpool: Ignore this, just noticed the other PR to correct it.

PR #140, opened 2016-10-05
force suspendDataStore to initialise prior...

moloko:
... to any attempt to store the user's chosen language - should fix https://github.com/adaptlearning/adapt-contrib-languagePicker/issues/18

remove workaround from adapt-offlineStorage-scorm.js as this should no longer be necessary


moloko:
@oliverfoster would appreciate you having a quick look at this to make sure I haven't done something stupid!

oliverfoster:
+1

PR #141, opened 2016-10-18
remove dependency on specific version of framework

moloko:
as per https://github.com/adaptlearning/adapt_framework/issues/1290

also sort out indentation

### DIFF
--- a/js/adapt-contrib-spoor.js
+++ b/js/adapt-contrib-spoor.js
@@ -20,7 +20,9 @@ define([
 
     onConfigLoaded: function() {
       if (!this.checkConfig()) {
-        Adapt.offlineStorage.setReadyStatus();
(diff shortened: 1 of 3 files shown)


js/adapt-contrib-spoor.js
```diff
@@ -48,9 +48,7 @@ define([
         },
 
         checkConfig: function() {
-            this._config = Adapt.config.has('_spoor') 
-                ? Adapt.config.get('_spoor')
-                : false;
+            this._config = Adapt.config.has('_spoor') ? Adapt.config.get('_spoor') : false;
```
oliverfoster: ```Adapt.config.get('_spoor') || false;`` < would be sufficient, no need to change it though

moloko: ta, I'll sort that out next time there's a requirement to change spoor

PR #142, opened 2016-11-11
Added warning to debug log if using offline API.

chris-steele:
Will help quickly determine if user has linked to incorrect launch file when deploying to LMS. This happened recently - a client used an uploader tool which clearly did not abide the settings in the imsmanifest.xml. index.html was selected as the launch file rather than index_lms.html and therefore no data was reported to the LMS.

**Expected usage**: tracking problems reported on LMS. Enable debugging in config, redeploy and first line of debug log will reveal if the appropriate tracking API is being used.

### DIFF
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -53,6 +53,9 @@ define (function(require) {
         
 		if (window.__debug)
 			this.showDebugWindow();
+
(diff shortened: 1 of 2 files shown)


oliverfoster:
+1

moloko:
Good idea. Also occurs to me that it would be good to have a keyboard shortcut to show the SCORM debugger...

chris-steele:
Absolutely - I'll look into this.

PR #143, opened 2016-11-14
Added keyboard shortcut to invoke Spoor logger

chris-steele:
Hold keys 'd', 'e', 'v' together to invoke log window. The included jquery.keycombo extension could perhaps go to core libs?


chris-steele:
Hmm - is that when running from a web server?

moloko:
yes - running from LMS

chris-steele:
OK that should be fixed now

moloko:
yep, works great for me in FF now

since there's nothing else in Adapt Core that requires keyboard shortcuts, it's hard to argue for putting jquery.keycombo in core. But perhaps it should be moved to the extension's 'libraries' folder in line with [this](https://github.com/adaptlearning/adapt-contrib-trickle/pull/73)?

moloko:
great stuff, now if you could just implement a 'tap d-e-v out in morse code' option for touch devices that would be great ;-)

chris-steele:
Haha, nice idea. I wonder, perhaps, for simplicity we could watch for a gesture like:

1. single finger touch (held)
2. second finger taps three times

Simple to use, unlikely to be triggered accidentally. What do you think?

moloko:
sorry @chris-steele you do need to change the framework version in README as well, just noticed that this has been done in other plugins (well, Slider)

PR #144, opened 2016-11-16
Typo Fix

sarveshwar-gavhane:
README is not upto date as example.json & properties.schema for _requireCourseCompleted


tomgreenfield:
Strictly speaking, is the default actually truthy in the code?

It may be misleading to say that the default is `true` just because that's what the authoring tool assigns. It infers that if a framework-only user was to leave the property out, it would default to `true`.

sarveshwar-gavhane:
oh i got confused in example.json & properties.schema and README

PR #145, opened 2016-11-29
Added support for IE 8 shim as defined in framework issue #1335

brian-learningpool:
This is dependent upon https://github.com/adaptlearning/adapt_framework/pull/1346.

Don't merge this until that goes in and the framework has been bumped.


brian-learningpool:
This is no longer required, as https://github.com/adaptlearning/adapt_framework/pull/1346 moves loading of the shim to `scriptLoader.js`.

PR #146, opened 2016-11-29
Resolves #1329, adds support for dynamic props in imsmanifest.xml

brian-learningpool:
The 'title' and 'description' are now passed through to their placeholders in imsmanifest.xml.


moloko:
It would be great to be able to replace all instances of `"adapt_"` in the manifest with some kind of unique id, strictly speaking some of these IDs are meant to be unique to the SCO on the LMS.

Is this possible and is there any suitable 'courseid' generated by the AT that could be used?

brian-learningpool:
Hi @moloko, yes, the `_id` value as held in MongoDB is unique to the authoring tool instance which created it.

Would that do?

moloko:
sounds ideal

brian-learningpool:
@moloko, having just had a chance to get back to this there's a slight snag.  While we do have unique `_id` values from the authoring tool, the one for the course model is actually replaced with the string "course". 

Can we raise this as a separate issue if needs be?  A timestamp is another option.

moloko:
Yes let's deal with that separately, getting this stuff in is far more important

PR #153, opened 2017-03-30
schema updates

moloko:
adds `_shouldRecordInteractions` into the schema to fix [#1494](https://github.com/adaptlearning/adapt_framework/issues/1494) 

also switch drop downs to checkboxes since it seems daft to make the user click twice to change the selection

amends to help text, spacing

### DIFF
--- a/properties.schema
+++ b/properties.schema
@@ -23,7 +23,7 @@
                   "title": "Is Enabled",
                   "inputType": "Checkbox",
                   "validators": [],
-                  "help": "If checked, the plugin will try to find and track to a SCORM conformant LMS. Uncheck to switch off SCORM tracking."
+                  "help": "If enabled, the plugin will try to find and track to a SCORM conformant LMS. Uncheck to switch off SCORM tracking."
                 },
                 "_tracking": {
                   "type": "object",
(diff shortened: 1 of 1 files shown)


properties.schema
```diff
@@ -1,70 +1,79 @@
...
-                  "type":"boolean",
-                  "required":true,
-                  "default": "true",
-                  "title":"Is Enabled",
-                  "inputType": {"type": "Boolean", "options": [true, false]},
+                  "type": "boolean",
+                  "required": true,
+                  "default": true,
+                  "title": "Is Enabled",
+                  "inputType": "Checkbox",
```
brian-learningpool: Replacing the select with a checkbox makes sense but I'm conscious that similar properties on other plugins still use <select> boxes and this introduces some inconsistency.

moloko: well, there is [a ticket to make them consistent](https://github.com/adaptlearning/adapt_authoring/issues/1380) that has been kicking around for ages so maybe this is a good opportunity to make that change?

moloko: also worth noting Helen's comment here https://github.com/adaptlearning/adapt_authoring/issues/1324

moloko: I would be happy to make this same change across all core schema plugins to make them consistent... just say the word!

oliverfoster: Kumquat!!?

moloko: @oliverfoster sorry that's not the word

I think you'll find the bird is the word... have you heard about the bird?

brian-learningpool: Hi @moloko, yes, I think this is a good opportunity to do that. I've updated the linked ticket just to keep track of it.

PR #154, opened 2017-04-21
issues-1535: remove cookie url encoding

oliverfoster:
https://github.com/adaptlearning/adapt_framework/issues/1535


moloko:
need to do a bit more testing but initially looks good

oliverfoster:
👍

PR #155, opened 2017-05-02
have spoor overwrite the hard-coded 'learner info' data with real data

moloko:
part of https://github.com/adaptlearning/adapt_framework/issues/1040


moloko:
> I know it's not always ln, fn

Well, in SCORM 1.2 it _should_ always be in that format:
student_name
However SCORM 2004 doesn't seem to have the same requirement.

PR #158, opened 2017-06-27
Implemented detection of when storage limit exceeded

chris-steele:
Fixes https://github.com/adaptlearning/adapt_framework/issues/1619.

N.B. respects `_spoor._advancedSettings._suppressErrors` if set (using `console.error` in preference to `alert`).

### DIFF
--- a/required/offline_API_wrapper.js
+++ b/required/offline_API_wrapper.js
@@ -10,6 +10,19 @@ function createResetButton(API) {
 	});
 }
 
+function storageWarning() {
+	var Adapt;
+	var notificationMethod = alert;
+	this.__storageWarningTimeoutId = null;
+	if (require) Adapt = require('coreJS/adapt');
+	if (Adapt && Adapt.config && Adapt.config.has('_spoor')) {
+		if (Adapt.config.get('_spoor')._advancedSettings &&
+			Adapt.config.get('_spoor')._advancedSettings._suppressErrors === true) {
+			notificationMethod = console.error;
+		}
+	}
+	notificationMethod('Warning: possible cookie storage limit exceeded - tracking may malfunction');
+}
 
 var API = {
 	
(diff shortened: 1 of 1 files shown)


required/offline_API_wrapper.js
```diff
@@ -68,7 +81,18 @@ var API = {
...
 		if (window.ISCOOKIELMS === false) return;
 		if (!force && API.cookie("_spoor") === undefined) return;
-		API.cookie("_spoor", JSON.stringify(this.data));
+
+		var stringified = JSON.stringify(this.data);
+
+		API.cookie("_spoor", stringified);
+
+		// a length mismatch will most likely indicate cookie storage limit exceeded
+		if (API.cookie("_spoor").length != stringified.length) {
```
oliverfoster: two opportunities to return early here, invert the logic and return instead of nesting

oliverfoster: sorry to steal your thunder @tomgreenfield

tomgreenfield: You're not wrong, @oliverfoster! Something like this?

```js
function storageWarning() {
	this.__storageWarningTimeoutId = null;
	var Adapt = require ? require('coreJS/adapt') : null;
	if (!Adapt) return;
	var spoor = Adapt.config ? Adapt.config.get('_spoor') : null;
	if (!spoor) return;
	var settings = config._advancedSettings;
	var notificationMethod = alert;
	if (settings && settings._suppressErrors) notificationMethod = console.error;
	notificationMethod('Warning: possible cookie storage limit exceeded - tracking may malfunction');
}
```

The limited-whitespace thing feels weird.

tomgreenfield: Hang on, not sure returning early is entirely correct since the error presents anyway?

chris-steele: @tomgreenfield no, you were correct, we still want to show the error. The if statements are just concerned with looking for `_suppressErrors` and if set to true we use `console.error` rather than the more in your face `alert`

chris-steele: I know the white space is, well, missing, but this was done in the spirit of `createResetButton`. That's no great justification I know.

moloko:
I've installed this in 403291 if anyone at Kineo wants to give it a try...

PR #163, opened 2017-09-12
add support for storing the user's choice of language

moloko:
in the appropriate SCORM data element (where supported) - see https://github.com/adaptlearning/adapt_framework/issues/1404

### DIFF
--- a/js/adapt-offlineStorage-scorm.js
+++ b/js/adapt-offlineStorage-scorm.js
@@ -86,6 +86,12 @@ define([
 				case "student":
 				case "learnerinfo":
 					return false;// these properties are read-only
+				case "lang":
+					scorm.setLanguage(value);
+					// fall-through so that lang gets stored in suspend_data as well:
+					// because in SCORM 1.2 cmi.student_preference.language is an optional data element
+					// so we can't rely on the LMS having support for it. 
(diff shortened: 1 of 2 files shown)


moloko:
I have tested this on:
* Moodle
* Saba Content Lab
* SCORM Cloud (as both SCORM 1.2 and SCORM 2004 4th edition)

Unfortunately all of those have support for `cmi.student_preference.language` - so I'm unable to test that it fails gracefully when that data element isn't supported... it should work fine, I'd just prefer to know for sure!

If anyone has access to an LMS that they know doesn't support `cmi.student_preference.language` I can share a test SCO with you to try out...

oliverfoster:
is there a sister pr for the languagePicker?

moloko:
@oliverfoster no because it already does that - though now you mention it I can't for the life of me figure out **how** it is doing it; I assumed there was a call to `Adapt.offlineStorage.set("lang", newlang);` somewhere in languagePicker but apparently not....! time for some forensic debugging

moloko:
aha: https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/app.js#L228

oliverfoster:
if it's not supported on the LMS the language picker won't save it's language in suspend data any more?

oliverfoster:
ignore me, i just read your comment

PR #164, opened 2017-09-25
#1811 add spoor-specific trigger when data is ready

danielstorey:
Now plugins that depend on spoor data can listen specifically for when the data is ready

### DIFF
--- a/js/adapt-contrib-spoor.js
+++ b/js/adapt-contrib-spoor.js
@@ -46,7 +46,7 @@ define([
         onDataReady: function() {
             Adapt.trigger('plugin:beginWait');
             adaptStatefulSession.initialize(function() {
-                Adapt.trigger('plugin:endWait');
+                Adapt.trigger('plugin:endWait spoor:dataReady');
             });
         },
 


js/adapt-contrib-spoor.js
```diff
@@ -46,7 +46,7 @@ define([
         onDataReady: function() {
             Adapt.trigger('plugin:beginWait');
             adaptStatefulSession.initialize(function() {
-                Adapt.trigger('plugin:endWait');
+                Adapt.trigger('plugin:endWait spoor:dataReady');
```
oliverfoster: do we need a more generic solution for this in core?

danielstorey: Make offlineStorage:ready more reliable? So you don't need lines like [this](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/js/adapt.js#L57)

oliverfoster: this is both offlineStorage and completion restoration. i'm not sure if one global status for "data + completion is ready, continue with rendering the course" is enough now that xapi is around the corner and we'll have mixed environments, it may well be?

if we could fix the waiting mechanism first and have a conversation about the restoration process and the startup i'd be happy? i'm not sure adding another event call to the mix is going to make it any easier without a clear sense of direction.

brian-learningpool: I agree with @oliverfoster's comment.  I think we need to consider that one plugin only can be responsible for the course state, right now either Spoor or xAPI, so I don't think we should prefix the event with `spoor`.  I would also avoid adding another event with the `dataReady` name to avoid confusion with the existing ones.  

Really what we want to know is whether the state has been successfully restored or not.

brian-learningpool: @danielstorey, per the comment about `adapt:start` (here: https://github.com/adaptlearning/adapt_framework/issues/1811) is it OK to close this PR?

danielstorey: Yeah. Shut it down.

PR #167, opened 2017-11-10
 Resolves #1875, defaults _shouldStoreResponses to true 

brian-learningpool:
After some discussion it was agreed that this is a more sensible default.  Also updated the label and help text for the authoring tool for more clarity.

### DIFF
--- a/properties.schema
+++ b/properties.schema
@@ -60,17 +60,17 @@
                     "_shouldStoreResponses": {
                       "type": "boolean",
                       "required": false,
-                      "default": false,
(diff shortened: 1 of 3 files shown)


properties.schema
```diff
@@ -60,17 +60,17 @@
                     "_shouldStoreResponses": {
                       "type": "boolean",
                       "required": false,
-                      "default": false,
-                      "title": "Remember responses",
+                      "default": true,
+                      "title": "Store question state",
```
oliverfoster: plural?

PR #168, opened 2017-11-15
issue #1857 - set default value for _shouldStoreResponses to true

danielstorey:
Warn user if this setting is disabled


brian-learningpool:
See https://github.com/adaptlearning/adapt-contrib-spoor/pull/167

PR #172, opened 2018-03-07
Resolves #2001, in preparation for v3 Spoor now listens for 'tracking…

brian-learningpool:
…:complete'

This update to Spoor is intended for release with v3 of the framework.  The following properties have been removed:
- _requireCourseCompleted
- _requireAssessmentPassed

These are both replaced with corresponding properties on the framework's config.json, which is used to trigger the 'tracking:complete' event.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,11 +59,7 @@ The attributes listed below are used in *config.json* to configure **Spoor**, an
 
 **_spoor**: (object): The Spoor object that contains values for **_isEnabled**, **_tracking**, **_reporting**, and **_advancedSettings**.
  
->**_isEnabled** (boolean): Enables/disables the **Spoor** extension. If set to `true` (the default value), the plugin will try to connect to a SCORM conformant LMS when the course is launched via *index_lms.html*. If one is not available, a 'Could not connect to LMS' error message will be displayed. This error can be avoided during course development either by setting this to `false` or - more easily - by launching the course via *index.html* or *main.html*. This latter technique is also useful if you are developing a course that could be run either from an LMS or a regular web server.
-
->>**_requireCourseCompleted** (boolean): Determines whether the learner must complete all the components in the course before the course can be marked as finished in the LMS. Acceptable values are `true` or `false`. The default is `false`.    
-
->>**_requireAssessmentPassed** (boolean): Determines whether the user must pass the course assessment, not simply complete it, before the course can be marked as finished in the LMS. Acceptable values are `true` or `false`. The default is `false`.  If this attribute and `_requireCourseCompleted` are both set to `true`, the learner must pass the course assessment as well as complete all components in order for the course can be marked as finished in the LMS.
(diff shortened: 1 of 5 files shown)


js/adapt-stateful-session.js
```diff
@@ -125,12 +128,34 @@ define([
...
+				case COMPLETION_STATE.FAILED: {
+					if (!this._config._reporting._onAssessmentFailure) {
+						Adapt.log.warn("No value defined for '_onAssessmentFailure', so defaulting to '" + completionStatus + "'");
+					} else {
+						completionStatus = this._config._reporting._onAssessmentFailure;
+					}
+				}
+			}
+
+			Adapt.offlineStorage.set("status", completionData.status.asLowerCase);
```
moloko: you should be using `completionStatus` here not `completionData.status.asLowerCase`

js/adapt-stateful-session.js
```diff
@@ -195,33 +218,12 @@ define([
 		},
 
 		submitAssessmentFailed: function() {
-			if (this._config && this._config._reporting.hasOwnProperty("_onAssessmentFailure")) {
-				var onAssessmentFailure = this._config._reporting._onAssessmentFailure;
-				if (onAssessmentFailure === "") return;
-					
-				Adapt.offlineStorage.set("status", onAssessmentFailure);
+			if (this._config && this._config._reporting._onAssessmentFailure) {
```
moloko: I don't _think_ this function is needed any more as it's now covered by `onTrackingComplete`... just doing a bit of testing to make sure that's the case though

README.md
```diff
@@ -59,11 +59,7 @@ The attributes listed below are used in *config.json* to configure **Spoor**, an
...
-
->>**_requireAssessmentPassed** (boolean): Determines whether the user must pass the course assessment, not simply complete it, before the course can be marked as finished in the LMS. Acceptable values are `true` or `false`. The default is `false`.  If this attribute and `_requireCourseCompleted` are both set to `true`, the learner must pass the course assessment as well as complete all components in order for the course can be marked as finished in the LMS.
+>>**_isEnabled** (boolean): Enables/disables the **Spoor** extension. If set to `true` (the default value), the plugin will try to connect to a SCORM conformant LMS when the course is launched via *index_lms.html*. If one is not available, a 'Could not connect to LMS' error message will be displayed. This error can be avoided during course development either by setting this to `false` or - more easily - by launching the course via *index.html* or *main.html*. This latter technique is also useful if you are developing a course that could be run either from an LMS or a regular web server.ß
```
tomgreenfield: Could you remove the reference to main.html please?

brian-learningpool: Sure thing.

js/adapt-stateful-session.js
brian-learningpool: The `tracking:complete` event does indeed cause the correct status to be sent.  

I didn't take this function out though as it's bound to assessment completion.  If the assessment was at the start or the middle of the course, for example, it might be beneficial to have this written back to the LMS while the user continues progress on the course?  Also, since the `score` is written in the same chain of events I thought it best to leave this as is.

moloko: This definitely isn't needed anymore.

PR #173, opened 2018-03-08
Issue #1970, switches 'configModel:dataLoaded' for 'offlineStorage:pr…

brian-learningpool:
…epare'

This allows using other plugins to implement offlineStorage providers, e.g. xAPI.

### DIFF
--- a/js/adapt-contrib-spoor.js
+++ b/js/adapt-contrib-spoor.js
@@ -16,7 +16,9 @@ define([
         initialize: function() {
             this.listenToOnce(Adapt, {
                 'offlineStorage:prepare': this.onPrepareOfflineStorage,
-                'app:dataReady': this.onDataReady
+                'app:dataReady': function() {
+                    Adapt.wait.for(adaptStatefulSession.initialize.bind(adaptStatefulSession));
+                }
             });
         },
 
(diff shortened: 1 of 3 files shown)


js/adapt-contrib-spoor.js
```diff
@@ -14,15 +14,15 @@ define([
     //Session Begin
 
         initialize: function() {
-            this.listenToOnce(Adapt, "configModel:dataLoaded", this.onConfigLoaded);
-            this.listenToOnce(Adapt, "app:dataReady", this.onDataReady);
+            this.listenToOnce(Adapt, {
+                'offlineStorage:prepare': this.onPrepareOfflineStorage,
+                'app:dataReady': this.onDataReady
```
moloko: considering `onDataReady` is now a one-line, single-use method, I don't think it needs its own method any more, could you this change to:
```js
this.listenToOnce(Adapt, {
    "configModel:dataLoaded": this.onConfigLoaded,
    "app:dataReady": function() {
        Adapt.wait.for(adaptStatefulSession.initialize.bind(adaptStatefulSession));
    }
});
```
while you're here?

brian-learningpool: No problem.

PR #176, opened 2018-04-24
reverts changes made in #171

moloko:
fixes https://github.com/adaptlearning/adapt_framework/issues/2046 by letting the compatibility code that's already present in questionView handle redirecting to the model/view version of the function as needed (as we were doing before)

note that when we come to drop that compatibility code from questionView - which will be quite a way off yet! - all that we'll need to change in spoor is to amend `onQuestionRecordInteraction` to this:
```js
var questionModel = questionView.model;
var responseType = questionModel.getResponseType();

// if responseType doesn't contain any data, assume that the question component hasn't been set up for cmi.interaction tracking
if(_.isEmpty(responseType)) return;

var id = questionModel.get('_id');
var response = questionModel.getResponse();
var result = questionModel.isCorrect();
var latency = questionView.getLatency();

Adapt.offlineStorage.set("interaction", id, response, result, latency, responseType);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ Note that due to the data storage limitations of browser cookies, there is less 
 Currently (officially) only supports SCORM 1.2  
 
 ----------------------------
-**Version number:**  3.0.0   <a href="https://community.adaptlearning.org/" target="_blank"><img src="https://github.com/adaptlearning/documentation/blob/master/04_wiki_assets/plug-ins/images/adapt-logo-mrgn-lft.jpg" alt="adapt learning logo" align="right"></a> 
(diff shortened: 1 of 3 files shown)


brian-learningpool:
I guess we should think about adding a deprecation warning into the compatibility layer when all the core plugins have moved these methods to the model?

oliverfoster:
That was my hope at the time of writing @brian-learningpool 
The compatibility layer can also be disabled with a boolean flag. So should be quite easy to test removing it without actually removing it. Or turn the compatibility layer into something you can control from the config.json

PR #177, opened 2018-07-05
issue/#1376 scorm error

chris-steele:
[#1376](https://github.com/adaptlearning/adapt_framework/issues/1376)

### DIFF
--- a/js/adapt-stateful-session.js
+++ b/js/adapt-stateful-session.js
@@ -1,9 +1,10 @@
 define([
 	'core/js/adapt',
+	'./scorm',
 	'./serializers/default',
 	'./serializers/questions',
 	'core/js/enums/completionStateEnum'
-], function(Adapt, serializer, questions, COMPLETION_STATE) {
+], function(Adapt, scorm, serializer, questions, COMPLETION_STATE) {
 
 	// Implements Adapt session statefulness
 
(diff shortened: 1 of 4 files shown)


chris-steele:
This change means that for matching and choice activities (matching/(g)mcq) if a numeric identifier is used (which _is_ the case) then identifiers 10-35 are mapped to a-z when using SCORM 1.2 to maintain adherence to the spec.

js/adapt-stateful-session.js
```diff
@@ -118,6 +119,50 @@ define([
...
+		/*
+		* SCORM 1.2 requires that the identifiers in cmi.interactions.n.student_response for choice and matching activities be a character from [0-9a-z].
+		* When numeric identifiers are used this function attempts to map identifiers 10 to 35 to [a-z]. Resolves issues/1376.
+		*/
+		_checkResponse: function(response, responseType) {
+			if (!response) return response;
+			if (responseType != 'choice' && responseType != 'matching') return response;
+
+			var zero = '0',
+				nine = '9';
```
oliverfoster: mix of variable list comma vs line end

js/adapt-stateful-session.js
```diff
@@ -118,6 +119,50 @@ define([
...
+				if (r.length == 1 && r >= zero && r <= nine) return r;
+
+				// if [a-z] then ok
+				if (r.length == 1 && r >= a && r <= z) return r;
+
+				// try to map integers 10-35 to [a-z]
+				i = parseInt(r);
+				if (isNaN(i) || i < 10 || i > 35) throw 'Numeric choice/matching response elements must use a value from 0 to 35 in SCORM 1.2';
+
+				return String.fromCharCode(base + i - 10); // 10 maps to 'a', 11 maps to 'b', ..., 35 maps to 'z'
```
oliverfoster: You can use `Number(i).toString(36)` instead to convert. It's base 36 instead of base 10.

js/adapt-stateful-session.js
```diff
@@ -118,6 +119,49 @@ define([
...
+		* When numeric identifiers are used this function attempts to map identifiers 10 to 35 to [a-z]. Resolves issues/1376.
+		*/
+		_checkResponse: function(response, responseType) {
+			if (!response) return response;
+			if (responseType != 'choice' && responseType != 'matching') return response;
+
+			var zero = '0',
+				nine = '9',
+				a = 'a',
+				z = 'z';
```
brian-learningpool: Is it necessary to actually declare variables for these?

Couldn't they be referenced inline as the strings '0', '9', etc.?

Also, not sure about the single var pattern going on here.

chris-steele:
@moloko I did consider the wrapper first, but I think the responsibility needs to be further up the chain towards the components creating the interaction data. Rather than put the same code in two places (mcq and matching) I settled for spoor. Also an error is thrown for any choice/matching activity passing in bad data and the wrapper at present does not throw any errors - so I didn't want to change this.

chris-steele:
On review I think it probably does sit better at one end of the chain rather than somewhere in the middle. Let's do the processing and error handling in the wrapper. We can also then take out the dependency I introduced to adapt-stateful-session.

js/adapt-stateful-session.js
chris-steele: Resolved

js/adapt-stateful-session.js
chris-steele: Resolved

js/adapt-stateful-session.js
chris-steele: Resolved

chris-steele:
@danielstorey @brian-learningpool @moloko @oliverfoster thanks guys will merge shortly unless anyone wants to defer it for additional review?

moloko:
@chris-steele assuming you've given it a good old test I'm happy for you to go ahead and merge, this looks good.

PR #181, opened 2019-01-08
SCORM Objectives support

chris-steele:
**N.B. this functionality is being introduced into the legacy version of spoor first. If successful it will be introduced into master.**

Objectives are implemented as being *completion-based* or *event-based*. A list of one or more Adapt models is given for a completion-based objective. As these models are completed the status of the objective changes. The functionality for completion-based objectives is built into spoor. Event-based objectives allow the state of the objective to be controlled externally (e.g. by a third party plugin). For event-based objectives it is the responsibility of the implementer to trigger a given event to request that the objective state should change.

## Objective types
Objectives can be completion-based or event-based. A completion-based objective changes state as the given content models (components, blocks, articles, content objects) are completed. State changes for event-based objectives are handled externally; allowing the behaviour of the objective to be implemented arbitrarily. Specify a completion-based objective using the property `_content` and assigning to it an array of content model IDs. An event-based objective is specified by assigning the name of the event to the `_event` property.

## Objective identifiers
Objective IDs can be given directly in `config.json`. Alternatively, objective IDs can be given by paths. A path uses standard dot and array notation. The path is relative to the global `Adapt` object. An example use of paths could be localisation; where the localised identifiers are stored with the associated JSON (e.g. `contentObjects.json`).

### Example objective ID paths
Prefix with `@@` to specify a path and use dot and array notation. All paths are relative to the global `Adapt` object. For example:
```
@@course._objectives.objective1._id
@@course._objectives[0]._id
@@contentObjects._byAdaptID[co-05][0]._objectiveId
```
## Objective states
On first run of a course an objective can be optionally initialised with the state `not attempted`. Specify whether this should happen with the property `_initialiseWithNotAttempted`. How the objective state changes depends on the type of objective: for completion-based objectives the state will change as the content models are completed. The state of event-based objectives is controlled externally; typically this will be from a custom plugin. When an objective enters a new state it can be locked into this state to prevent future changes. Such states are described as immutable. By default the completed state is immutable. Specify which states are immutable by using the `_immutableStates` property.

## Notes
Storage of interactions by the offline API was removed by this change. To allow objectives (and interactions) to be stored by the offline API for testing purposes there are two new configuration options in spoor: `_debugInteractions` and `_debugObjectives`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ The attributes listed below are used in *config.json* to configure **Spoor**, an
 
 #### Attributes
 
-**_spoor**: (object): The Spoor object that contains values for **_isEnabled**, **_tracking**, **_reporting**, and **_advancedSettings**.
+**_spoor**: (object): The Spoor object that contains values for **_isEnabled**, **_tracking**, **_reporting**, **_advancedSettings**, and **_objectives**.
  
 >**_isEnabled** (boolean): Enables/disables the **Spoor** extension. If set to `true` (the default value), the plugin will try to connect to a SCORM conformant LMS when the course is launched via *index_lms.html*. If one is not available, a 'Could not connect to LMS' error message will be displayed. This error can be avoided during course development either by setting this to `false` or - more easily - by launching the course via *index.html* or *main.html*. This latter technique is also useful if you are developing a course that could be run either from an LMS or a regular web server.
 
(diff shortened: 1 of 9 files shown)


oliverfoster:
closing as needs redoing

PR #182, opened 2019-01-22
Resolves #2247, adds support for editing 'identifier' attribute

brian-learningpool:
Some LMSs are picky with regards to the identifier used in imsmanifest.xml so
this is now a configurable option, defaulted to ~'adapt_scorm'~ 'adapt_manifest' (as is).

### DIFF
--- a/properties.schema
+++ b/properties.schema
@@ -167,6 +167,15 @@
                       "inputType": "Checkbox",
                       "validators": [],
                       "help": "If enabled, a 'commit' call will be made whenever the course window is hidden/minimised. Requires a browser that supports the 'visibilitychange' event."
+                    },
+                    "_manifestIdentifier": {
+                      "type": "string",
+                      "required": true,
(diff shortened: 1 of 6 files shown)


brian-learningpool:
Noted.  I'll update the README and the bower.json whenever v3 has the corresponding fallback for the missing attribute.

properties.schema
```diff
@@ -167,6 +167,15 @@
                       "inputType": "Checkbox",
                       "validators": [],
                       "help": "If enabled, a 'commit' call will be made whenever the course window is hidden/minimised. Requires a browser that supports the 'visibilitychange' event."
+                    },
+                    "_manifestIdentifier": {
+                      "type": "string",
+                      "required": true,
+                      "default": "adapt_scorm",
```
moloko: So as to maintain backwards compatibility, I think this should be:
```suggestion
                      "default": "adapt_manifest",
```

PR #183, opened 2019-03-29
update to use 2 spaces for indentation

nilslp:
Spoor converted to use 2 spaces for indentation.

I haven't touched any of the library or required files as these are usually copy/paste replacements.

### DIFF
--- a/js/adapt-stateful-session.js
+++ b/js/adapt-stateful-session.js
@@ -1,4 +1,3 @@
-
 define([
   'core/js/adapt',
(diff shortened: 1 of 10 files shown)


js/adapt-stateful-session.js
```diff
@@ -1,219 +1,220 @@
+
```
oliverfoster: Can you undo this?

nilslp:
@oliverfoster can you take another look please.

PR #184, opened 2019-06-19
Define Exit Status in Advanced Settings

canstudios-matth:
Fixes [#2458](https://github.com/adaptlearning/adapt_framework/issues/2458)

- Add fields to advanced settings that allow the user to define exit codes when course is completed, and when it is incomplete
- Support for SCORM 1.2 and 2004

### DIFF
--- a/js/adapt-contrib-spoor.js
+++ b/js/adapt-contrib-spoor.js
@@ -59,31 +59,31 @@ define([
 
                 scorm.setVersion(settings._scormVersion || "1.2");
 
-                if(settings.hasOwnProperty("_suppressErrors")) {
+                if(settings._suppressErrors) {
                     scorm.suppressErrors = settings._suppressErrors;
                 }
 
-                if(settings.hasOwnProperty("_commitOnStatusChange")) {
+                if(settings._commitOnStatusChange) {
                     scorm.commitOnStatusChange = settings._commitOnStatusChange;
                 }
 
-                if(settings.hasOwnProperty("_timedCommitFrequency")) {
+                if(_.isFinite(settings._timedCommitFrequency)) {
                     scorm.timedCommitFrequency = settings._timedCommitFrequency;
                 }
 
-                if(settings.hasOwnProperty("_maxCommitRetries")) {
+                if(_.isFinite(settings._maxCommitRetries)) {
                     scorm.maxCommitRetries = settings._maxCommitRetries;
                 }
 
-                if(settings.hasOwnProperty("_commitRetryDelay")) {
+                if(_.isFinite(settings._commitRetryDelay)) {
                     scorm.commitRetryDelay = settings._commitRetryDelay;
                 }
 
-                if (settings.hasOwnProperty("_exitStateIfIncomplete")) {
+                if ("_exitStateIfIncomplete" in settings) {
                     scorm.exitStateIfIncomplete = settings._exitStateIfIncomplete;
                 }
 
-                if (settings.hasOwnProperty("_exitStateIfComplete")) {
+                if ("_exitStateIfComplete" in settings) {
                     scorm.exitStateIfComplete = settings._exitStateIfComplete;
                 }
             } else {
--- a/js/adapt-contrib-spoor.js
+++ b/js/adapt-contrib-spoor.js
@@ -78,6 +78,22 @@ define([
                 if(settings.hasOwnProperty("_commitRetryDelay")) {
                     scorm.commitRetryDelay = settings._commitRetryDelay;
                 }
+
+                if (settings.hasOwnProperty("_exitStateIfIncomplete")) {
+                    if (settings._exitStateIfIncomplete === "'' (empty string)") {
+                        scorm.exitStateIfIncomplete = "";
+                    } else {
+                        scorm.exitStateIfIncomplete = settings._exitStateIfIncomplete;
+                    }
+                }
+
+                if (settings.hasOwnProperty("_exitStateIfComplete")) {
+                    if (settings._exitStateIfComplete === "'' (empty string)") {
+                        scorm.exitStateIfComplete = "";
+                    } else {
+                        scorm.exitStateIfComplete = settings._exitStateIfComplete;
+                    }
+                }
             } else {
                 /**
                 * force use of SCORM 1.2 by default - some LMSes (SABA/Kallidus for instance) present both APIs to the SCO and, if given the choice,
--- a/js/adapt-contrib-spoor.js
+++ b/js/adapt-contrib-spoor.js
@@ -80,19 +80,11 @@ define([
                 }
 
                 if (settings.hasOwnProperty("_exitStateIfIncomplete")) {
-                    if (settings._exitStateIfIncomplete === "'' (empty string)") {
-                        scorm.exitStateIfIncomplete = "";
-                    } else {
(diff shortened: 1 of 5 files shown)


js/scorm/wrapper.js
```diff
@@ -301,14 +306,31 @@ define ([
...
-                this.scorm.set("cmi.core.exit", "");
+
+                if (is_complete) {
+                    exit_value = this.exitStateIfComplete === "auto" ? "" : this.exitStateIfComplete;
+                } else {
+                    exit_value = this.exitStateIfIncomplete === "auto" ? "" : this.exitStateIfIncomplete;
+                }
             }
 
+            this.scorm.set(exit_property, exit_value);
```
moloko: it feels a little clumsy to have this here, I think it would actually be clearer to have it set inside the if...else code block above (similar to how `session_time` is handled), then you could drop the `exit_property` variable.

js/scorm/wrapper.js
```diff
@@ -301,14 +306,31 @@ define ([
 
             this.endTime = new Date();
 
+            var exit_property = "cmi.core.exit";
+            var exit_value = "";
+            var is_complete = this.scorm.data.completionStatus === "completed" || this.scorm.data.completionStatus === "passed";
```
moloko: `isComplete` would be more in keeping with the general style used in Adapt

js/adapt-contrib-spoor.js
```diff
@@ -78,6 +78,14 @@ define([
                 if(settings.hasOwnProperty("_commitRetryDelay")) {
                     scorm.commitRetryDelay = settings._commitRetryDelay;
                 }
+
+                if (settings.hasOwnProperty("_exitStateIfIncomplete")) {
```
moloko: this is probably a good opportunity to stop abusing `hasOwnProperty` in this way! 

@oliverfoster sorry I know you've covered this before but what's the best way to check for a property when it could have a legitimately 'falsy' value like `""` or `0`? Would something like `if ('_exitStateIfIncomplete' in settings)` be more appropriate?

js/scorm/wrapper.js
```diff
@@ -301,14 +306,31 @@ define ([
 
             this.endTime = new Date();
 
+            var exit_property = "cmi.core.exit";
+            var exit_value = "";
```
moloko: `exitState` would be more in keeping with the Adapt naming convention and the names you've used for this elsewhere e.g. `exitStateIfComplete `

js/adapt-contrib-spoor.js
oliverfoster:
```js
// when property value is checked and is not a falsy
if (object.attributeName) { 
 // value can be anything but: not "", 0, null, undefined or NaN
} 

// when instance property should exist and not be inherited from the prototype chain
if (object.hasOwnPropert(attributeName)) {
  // property must have been set on the instance
  // value can be anything
}

// when the property should exist and can be inherited
if (attributeName in object) {
  // property can have been inherited from the prototype chain
  // value can be anything
}
```
### References

* [in operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in)
* [hasOwnProperty function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty)
* [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)
* [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy)

moloko: Thanks @oliverfoster, very helpful!

@canstudios-matth the checks for `_suppressErrors` and `_commitOnStatusChange` can be changed to `if (settings._suppressErrors)` and `if (settings._commitOnStatusChange)`. 

As properties `_timedCommitFrequency`, `_maxCommitRetries` & `_commitRetryDelay` should all be numerical (with `0` as a valid value), some sort of check for a numerical value would be good, perhaps `_.isFinite`?

As the new settings can have `""` as a valid value they can either be left as `hasOwnProperty` or changed to use the `in` operator, as you prefer.

js/scorm/wrapper.js
canstudios-matth: Is snakeCase the standard? If so, I can change the other new variables.

js/adapt-contrib-spoor.js
oliverfoster:
```js
if (object.property || object.property === 0) {

}

if (_.isNumber(object.property)) {

}
```

Either of the above is good. JSON can't represent non-finite numbers. Positive and negative floating point integers and scientific notation only I believe. Such that:
```js
JSON.parse("Infinity"); // throws an error
JSON.parse("3.141e7"); // returns 31410000
JSON.stringify(Infinity); // returns null
```

#### References:
https://underscorejs.org/#isNumber
https://underscorejs.org/#isFinite
https://www.json.org/

js/scorm/wrapper.js
```diff
@@ -50,6 +59,8 @@ define ([
 
         this.logger = Logger.getInstance();
         this.scorm = pipwerks.SCORM;
+        // Prevent the Pipwerks SCORM API wrapper's handling of the exit status
```
moloko: Could you change to this style to match the other comments?
```
/**
 * Prevent the Pipwerks SCORM API from handling the exit status
 */

File: js/adapt-contrib-spoor.js (hunk 3 above)
Comment (moloko) [reply js/adapt-contrib-spoor.js]: @oliverfoster my main reason for suggesting `_.isFinite` rather than `_.isNumber` is that the latter will return `true` for `NaN` whereas the former will not. But perhaps it doesn't really matter.

File: js/scorm/wrapper.js [hunk 6]
```diff
@@ -716,5 +722,21 @@ define ([
         return response.join(',');
     };
 
+    ScormWrapper.prototype.getExitState = function() {
+        var completionStatus = this.scorm.data.completionStatus;
+        var isComplete = completionStatus === 'completed' || completionStatus === 'passed';
+        var exitState = isComplete ? this.exitStateIfComplete : this.exitStateIfIncomplete;
+
+        switch (exitState) {
```
moloko: I don't see how `exitState` could be `undefined` so we can drop the `switch` as per @tomgreenfield 's suggestion and do something like this instead:
```js
if (exitState !== 'auto') return exitState;

if (!this.isSCORM2004()) return '';

return isComplete ? 'normal' : 'suspend';
```

moloko:
thanks @canstudios-matth this is looking pretty good, I'd still like to do some testing on it though - I'll try and find the time to do that tomorrow. there's quite a lot of combinations of settings that will need checking!

js/scorm/wrapper.js
```diff
@@ -716,5 +722,17 @@ define ([
         return response.join(',');
     };
 
+    ScormWrapper.prototype.getExitState = function() {
+        var completionStatus = this.scorm.data.completionStatus;
+        var isIncomplete = completionStatus === 'incomplete' || completionStatus === 'not attempted';
+        var exitState = isComplete ? this.exitStateIfComplete : this.exitStateIfIncomplete;
```
moloko: You need to amend this line as a result of the changes you made in 
bc3b19c

moloko:
Oh, and just to add I haven't done any testing in the authoring tool so have only 👀 your additions to properties.schema

PR #186, opened 2019-10-03
Issue 2537 Setting context for the handleError call

cahirodoherty-learningpool:
Fixes https://github.com/adaptlearning/adapt_framework/issues/2537

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "adapt-contrib-spoor",
-  "version": "3.3.0",
+  "version": "3.3.1",
   "framework": ">=3.5.0",
   "homepage": "https://github.com/adaptlearning/adapt-contrib-spoor",
   "issues": "https://github.com/adaptlearning/adapt_framework/issues/new",
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -691,6 +691,8 @@ define ([
 
         response = response.split(/,|#/);
 
+        var self = this;
+
         if (responseType == 'choice') {
             response = response.map(checkIdentifier);
         } else {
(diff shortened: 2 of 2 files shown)


moloko:
> 
> 
> can we get rid of self?
> 
> ```
> if (responseType == 'choice') {
>     response = response.map(checkIdentifier.bind(this));
> } else {
>     response = response.map(function(r) {
>         var identifiers = r.split('.');
>         return checkIdentifier(identifiers[0]) + '.' + checkIdentifier(identifiers[1]);
>     }.bind(this));
> }
> ```

Would that work though? I'd have thought you'd have to do something more like this:
```js
response = response.map(function(r) {
  var identifiers = r.split('.');
  return checkIdentifier(identifiers[0]).bind(this) + '.' + checkIdentifier(identifiers[1]).bind(this);
});
```
which, if correct, is starting to look a little clumsy

NayanKhedkar:
> Would that work though? I'd have thought you'd have to do something more like this:
> 
> ```js
> response = response.map(function(r) {
>   var identifiers = r.split('.');
>   return checkIdentifier(identifiers[0]).bind(this) + '.' + checkIdentifier(identifiers[1]).bind(this);
> });
> ```
> 
> which, if correct, is starting to look a little clumsy

yes @moloko you are right

```
response = response.map(function(r) {
  var identifiers = r.split('.');
  return checkIdentifier.call(this,identifiers[0]) + '.' + checkIdentifier.call(this,identifiers[1]);
}.bind(this));
```
yeah its look a little clumsy. Drop It

moloko:
@cahirodoherty-learningpool before I merge can I just check that you tested this fixes the issue?

cahirodoherty-learningpool:
> @cahirodoherty-learningpool before I merge can I just check that you tested this fixes the issue?

Hi @moloko I did indeed, and it now displays the error message as expected, instead of falling over

PR #190, opened 2020-05-07
issue/2744 ES6 rewrite scormSuspendDataSerializer.js

olivermartinfoster:
[#2744](https://github.com/adaptlearning/adapt_framework/issues/2744)
* Enhanced compression significantly
* Added documentation
* Made much easier to read
* Supported in fw 5.3.0+

Standard course question responses suspend data:
```js
[[[0,7,0,0],[true,true,true,true,false],[false,false,false,true]],[[0,8,1,0],[true,true,true,true,true],[false,true,false]],[[0,9,1,0],[true,true,true,true,true],[65536]],[[0,10,0.33,0],[true,true,true,true,false],[0,3,1]],[[0,11,0,0],[true,false,true,true,false],[4]],[[0,12,0,0],[true,true,true,true,false],[true,true,true,true]],[[0,13,0,0],[true,true,true,true,false],[true,true,true,true,true]],[[0,15,0,0],[true,true,true,true,false],[true,true,true,true]],[[0,16,0,0],[true,true,true,true,false],[false,false,false,true]],[[0,17,0,0],[true,true,true,true,false],[true,true,true,true]]]
```

Using `SCORMSuspendData.serialize()`

Old algorithm output (169 characters):
`
"CIE6IFhAcADgx4KAFOiBYQIEA4MfCgSaogWECRAODHyZAAAQAAmohaUDAAAwAoQAiEMAADgx4SEAxTogWECwAODFgoRE6IFhAwADgx4KA9SiBYQNAA4MeDgx9OiBYQPAA4MeCgPXogmIABAAAA4MeCgBXogmIABEAAA4MeCgP"
`

New algorithm output (102 characters):
`
"lQ4nDgFeIEcTCEEFfGQ4mEJIK-H5gACAAA4noEKBQgK8PgyOJhCwCrB6DiYQwArxHjiYQ0ArxXxxMIeAV4jxxMIgAV4gRxMIiAV4jw"
`

We should be able to fit 39% more question responses into SCORM 1.2 suspend data.

Includes pr #192

PR #192, opened 2020-05-07
issue/2747 Added _attemptStates save and restore 

olivermartinfoster:
[#2747](https://github.com/adaptlearning/adapt_framework/issues/2747)

Requires [#2748](https://github.com/adaptlearning/adapt_framework/pull/2748) to framework for testing.
Built atop of #190, the old question state serialiser couldn't handle the more complicated data structure.

#### Changed
* All question states are saved all of the time, rather than just complete questions. All states would normally only be stored by the completion of the course.
* Questions with `_isResetOnRevisit = truthy` are saved as attempt states can be more useful here. The new serialiser is much more efficient and stable and is able to capture this new information and store all of the current and new data in a smaller suspend data footprint.
* `_trackingId` property can now indicate that the block shouldn't be saved / restored by setting to `null`, `undefined` or `<0` - this is useful for cloned models or runtime models which shouldn't expect to have a tracking id.
* Question states are saved when the question is submitted not when it is completed.
* `_attemptsLeft` will be restored across multiple sessions where no `_isResetOnRevisit` or other `model.reset` behaviour is called. This means a user cannot not get more question attempts than are specified without a model reset (like with `_isResetOnRevisit` or assessment question reset).

#### Added
* Save and restores `_attemptStates` attributes across sessions
* Added `_spoor._tracking._shouldStoreAttempts = false` to `config.json` to control when attempt states are saved and restored. Attempt states are not saved and restored by default.

Includes prs #193 and #194


js/serializers/questions.js
```diff
@@ -107,6 +113,23 @@ define([
...
 
             dataItem.push(userAnswer);
+          } else {
+            dataItem.push([]);
+          }
+
+          if (shouldStoreAttempts && hasAttemptStates) {
+            var attemptStates = isAttemptStatesArray ? component['_attemptStates'] : [component['_attemptStates']];
+
+            var invalidError = SCORMSuspendData.getInvalidTypeError(userAnswer);
```
tomgreenfield, line 123: Redeclared variable.

js/adapt-offlineStorage-scorm.js
```diff
@@ -13,6 +13,14 @@ define([
 
   Adapt.offlineStorage.initialize({
 
+    save: function() {
+      // listens to Adapt{tracking:save} event in adapt-stateful-session
```
moloko, line 17: out of curiosity, is there a specific reason for this to be triggered by the event rather than using the hard API? If so, it would be useful to add a brief explanation here

js/adapt-stateful-session.js
```diff
@@ -61,29 +61,41 @@ define([
...
 
+      if (sessionPairs.c) {
+        courseState = SCORMSuspendData.deserialize(sessionPairs.c);
+      }
+
       // Asynchronously restore block completion data because this has been known to be a choke-point resulting in IE8 freezes
-      if (sessionPairs.completion) {
-        serializer.deserialize(sessionPairs.completion, doSynchronousPart);
+      if (courseState) {
+        serializer.deserialize(courseState.slice(2).map(Number).map(String).join(''), doSynchronousPart);
```
moloko, line 80: > this has been known to be a choke-point resulting in IE8 freezes

Is there a reason we're still doing this even though we stopped supporting IE8 some time ago? If it's just because there are performance concerns with IE11 still, let's update the comment to make that clear.

js/serializers/questions.js
```diff
@@ -6,8 +6,7 @@ define([
   //Captures the completion status and user selections of the question components
   //Returns and parses a base64 style string
   var includes = {
-      "_isQuestionType": true,
-      "_isResetOnRevisit": false
+    "_isQuestionType": true
```
moloko, line 9:
```suggestion
    '_isQuestionType': true
```

js/serializers/questions.js
```diff
@@ -37,9 +36,11 @@ define([
     captureData: function() {
       var data = [];
 
-      var trackingIds = Adapt.blocks.pluck("_trackingId");
+      var trackingIds = Adapt.blocks.pluck("_trackingId").filter(id => (Number.isInteger(id) && id >= 0));
```
moloko, line 39:
```suggestion
      var trackingIds = Adapt.blocks.pluck('_trackingId').filter(id => (Number.isInteger(id) && id >= 0));
```

js/serializers/questions.js
```diff
@@ -75,6 +71,13 @@ define([
             isUserAnswerArray = false;
           }
 
+          var hasAttemptStates = (component['_attemptStates'] !== undefined);
+          var isAttemptStatesArray = (component['_attemptStates'] instanceof Array);
```
moloko, line 75:
```suggestion
          var isAttemptStatesArray = Array.isArray(component._attemptStates);
```
slightly neater?

js/serializers/questions.js
```diff
@@ -85,6 +88,9 @@ define([
           var booleanParameters = [
             hasUserAnswer,
             isUserAnswerArray,
+            hasAttemptStates,
+            isAttemptStatesArray,
+            component['_isComplete'],
```
moloko, line 93: It would be nice - but obviously by no means essential - to switch any instances of these that don't need to be written like this over to dot-notation.
```suggestion
            component._isComplete,
```

js/serializers/questions.js
```diff
@@ -95,18 +101,35 @@ define([
...
+            dataItem.push([]);
+          }
+
+          if (shouldStoreAttempts && hasAttemptStates) {
+            var attemptStates = isAttemptStatesArray ? component['_attemptStates'] : [component['_attemptStates']];
+
+            invalidError = SCORMSuspendData.getInvalidTypeError(userAnswer);
+
+            if (invalidError) {
+              console.log("Cannot store _attemptStates from component " + component._id + " as array is invalid", invalidError);
```
moloko, line 126:
```suggestion
              console.log(`Cannot store _attemptStates from component ${component._id} as array is invalid`, invalidError);
```

js/serializers/questions.js
```diff
@@ -145,16 +171,19 @@ define([
...
         var block = Adapt.blocks.findWhere({_trackingId: trackingId});
         var components = block.getChildren();
         components = components.where(includes);
         var component = components[blockLocation];
 
-        component.set("_isComplete", true);
+        component.set("_isComplete", isComplete);
         component.set("_isInteractionComplete", isInteractionComplete);
         component.set("_isSubmitted", isSubmitted);
         component.set("_score", score);
```
moloko, line 189:
```suggestion
        component.set({
          _isComplete: isComplete,
          _isInteractionComplete: isInteractionComplete,
          _isSubmitted: isSubmitted,
          _score: score
        });
```

properties.schema
```diff
@@ -48,6 +48,15 @@
...
                       "help": "If enabled, the user's responses to questions will be saved and restored in each session. Note: this setting should be enabled in the majority of cases."
                     },
+                    "_shouldStoreAttempts": {
+                      "type": "boolean",
+                      "required": false,
+                      "default": false,
+                      "title": "Store question attempt states",
+                      "inputType": "Checkbox",
+                      "validators": [],
+                      "help": "If enabled, the user's attempt states to questions will be saved and restored in each session. Note: this setting should be disabled in the majority of cases."
```
moloko, line 58: > Note: this setting should be disabled in the majority of cases.

Would be useful to have either here or in the README (and this setting needs adding to README too I think?) an explanation of when it might be useful to have this enabled.

js/serializers/questions.js
```diff
@@ -168,6 +197,12 @@ define([
           component.set("_userAnswer", userAnswer);
         }
 
+        if (shouldStoreAttempts && hasAttemptStates) {
+          var attemptStates = dataItem[3];
+          if (!isAttemptStatesArray) attemptStates = attemptStates[0];
+
+          component.set("_attemptStates", attemptStates);
```
moloko, line 204:
```suggestion
          component.set('_attemptStates', attemptStates);
```

js/serializers/questions.js
tomgreenfield: I half-offered to @oliverfoster that I'd refactor `questions.js` at same point; it definitely needs a bit of TLC.

js/serializers/questions.js
oliverfoster:
```js
        component.set({
          _isComplete,
          _isInteractionComplete,
          _isSubmitted,
          _score
        });
```
This is also possible now.

js/adapt-offlineStorage-scorm.js
oliverfoster: there was some ambiguous and currently out of grasp logic to it, i'll change. :+1:

js/adapt-stateful-session.js
oliverfoster: i didn't want to change it because it works and ie11 testing is a bit too big for this pr, will add a comment as such

properties.schema
oliverfoster: We don't have any use-cases for this at the moment. It will be for branching-style behaviour.

PR #195, opened 2020-06-09
issue/2790 Added commitOnAnyChange

oliverfoster:
[#2790](https://github.com/adaptlearning/adapt_framework/issues/2790)
* Added `_spoor._advancedSettings._commitOnAnyChange` to allow saving data to the LMS any time the SCORM data is changed.

This should help patch the current situation where Chrome isn't able to call `scorm.finish` properly as the synchronous XHR request is blocked.

Note that enabling this setting may have a large impact on the server and should only be used if absolutely necessary. An alternative is to use a lower Frequency value as it will have a lower to impact on the server.


moloko:
I have a couple of concerns about the implementation of this... firstly it does have the potential to rather overwhelm the LMS server if it has a lot of courses with a lot of learners. the other issue is that it still won't ensure that all the data is saved because some data only gets saved properly by the LMS when `LMSFinish` is successfully called.

I think it would be better to encourage people to include a close/exit button in the course and update [adapt-close](https://github.com/cgkineo/adapt-close/) so that it calls LMSFinish, which would just require a bit of modification [here](https://github.com/cgkineo/adapt-close/blob/c695842f63cad6937a5ae1efd0c3f01c6e4527a0/js/adapt-close.js#L46-L57)

That said, as this is a fairly minor change I've got no problem with adding it in but we could use a bit more info around the potential impact of enabling this.

oliverfoster:
> some data only gets saved properly by the LMS when LMSFinish is successfully called

If it `session_time` and `exit` or is there more to it?
https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/js/scorm/wrapper.js#L310-L316
https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/libraries/SCORM_API_wrapper.js#L379-L467

>  firstly it does have the potential to rather overwhelm the LMS server if it has a lot of courses with a lot of learners

Whilst it is a large amount more data, it is debounced so that values will only get set in batches. That would be a good example of something to add to the comments.

> I think it would be better to encourage people to include a close/exit button in the course and update adapt-close so that it calls LMSFinish, which would just require a bit of modification here

I concerned that requiring people to use alternative UI to close a window isn't going to work.

> That said, as this is a fairly minor change I've got no problem with adding it in but we could use a bit more info around the potential impact of enabling this.

AAT and readme? What did you want me to add?

README.md
```diff
@@ -80,6 +80,8 @@ The attributes listed below are used in *config.json* to configure **Spoor**, an
 
 >>**\_commitOnStatusChange** (boolean): Determines whether a "commit" call should be made automatically every time the SCORM *lesson_status* is changed. The default is `true`.  
 
+>>**\_commitOnAnyChange** (boolean): Determines whether a "commit" call should be made automatically every time any SCORM value is changed. The default is `false`. Note that setting **\_commitOnAnyChange** to `true` may have a large impact on the server and should only be used if absolutely necessary. An alternative is to use a lower **\_timedCommitFrequency** value as it will have a lower to impact on the server.
```
moloko, line 83:
```suggestion
>>**\_commitOnAnyChange** (boolean): Determines whether a "commit" call should be made automatically _every time_ any SCORM value is changed. The default is `false`. Setting **\_commitOnAnyChange** to `true` will disable 'timed commits'. **Note:** enabling this setting will make the course generate a lot more client-server traffic so you should only enable it if you are sure it is needed and, as it may have a detrimental impact on server performance, after careful load-testing. An alternative might be to first try setting a lower value for **\_timedCommitFrequency**.
```

example.json
```diff
@@ -18,6 +18,8 @@
             "_showDebugWindow": false,
             "_suppressErrors": false,
             "_commitOnStatusChange": true,
+            "_comment": "Note that enabling _commitOnAnyChange may have a large impact on the server and should only be used if absolutely necessary. An alternative is to use a lower _timedCommitFrequency value as it will have a lower to impact on the server.",
+            "_commitOnAnyChange": false,
```
moloko, line 22:
```suggestion
            "_commitOnAnyChange": false,
```
this is covered pretty comprehensively in the README and properties.schema, I don't think we need to repeat it a third time.

js/scorm/wrapper.js
```diff
@@ -13,6 +13,12 @@ define ([
      * whether to commit each time there's a change to lesson_status or not
      */
     this.commitOnStatusChange = true;
+    /**
+     * whether to commit each time there's a change to any value
+     * note that enabling commitOnAnyChange may have a large impact on the server and should only be used if absolutely necessary
+     * an alternative is to use a lower timedCommitFrequency value as it will have a lower to impact on the server
+     */
```
moloko, line 20:
```suggestion
    /**
     * whether to commit each time there's a change to any value
     */
```

properties.schema
```diff
@@ -124,6 +124,15 @@
...
                       "help": "If enabled, a 'commit' call will be made any time the lesson_status is changed."
                     },
+                    "_commitOnAnyChange": {
+                      "type": "boolean",
+                      "required": false,
+                      "default": true,
+                      "title": "Commit data on any change",
+                      "inputType": "Checkbox",
+                      "validators": [],
+                      "help": "If enabled, a 'commit' call will be made any time a value is changed. Note that enabling this setting may have a large impact on the server and should only be used if absolutely necessary. An alternative is to use a lower Frequency value as it will have a lower to impact on the server."
```
moloko, line 134:
```suggestion
                      "help": "If enabled, a 'commit' (save all data) call will be made any time a value is changed. Note that enabling this setting may have a detrimental impact on the server and should only be used if absolutely necessary and only after careful load testing. Consider trying a lower value for 'Frequency of automatic commits' before changing this setting."
```

oliverfoster:
Thanks

properties.schema
```diff
@@ -133,6 +133,15 @@
                       "validators": [],
                       "help": "If enabled, a 'commit' call will be made any time the lesson_status is changed."
                     },
+                    "_commitOnAnyChange": {
+                      "type": "boolean",
+                      "required": false,
+                      "default": true,
```
lc-alexanderbenesch, line 139: should be false

lc-alexanderbenesch (changes requested):

properties.schema
moloko: nice catch, thanks!

lc-alexanderbenesch (approved):
👍

PR #198, opened 2020-08-05
issue/1333 Added error abstraction

oliverfoster:
[#1333](https://github.com/adaptlearning/adapt_framework/issues/1333)
[#1238](https://github.com/adaptlearning/adapt_framework/issues/1238)

### Added
* Error objects with codes and error data
* Default messages and translatable overrides
* Used an `_isCancellable: false` notify to display `CLIENT_COULD_NOT_CONNECT` errors

image

### DIFF
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -1,7 +1,21 @@
 define([
+  'core/js/adapt',
   'libraries/SCORM_API_wrapper',
-  './logger'
-], function(pipwerks, Logger) {
+  './logger',
+  './error'
+], function(Adapt, pipwerks, Logger, ScormError) {
+
+  const {
+    CLIENT_COULD_NOT_CONNECT,
+    SERVER_STATUS_UNSUPPORTED,
+    CLIENT_STATUS_UNSUPPORTED,
+    CLIENT_COULD_NOT_COMMIT,
+    CLIENT_NOT_CONNECTED,
+    CLIENT_COULD_NOT_FINISH,
+    CLIENT_COULD_NOT_GET_PROPERTY,
+    CLIENT_COULD_NOT_SET_PROPERTY,
+    CLIENT_INVALID_CHOICE_VALUE
+  } = ScormError;
 
   /**
    * IMPORTANT: This wrapper uses the Pipwerks SCORM wrapper and should therefore support both SCORM 1.2 and 2004. Ensure any changes support both versions.
@@ -104,7 +118,7 @@ define([
 
         this.initTimedCommit();
       } else {
-        this.handleError('Course could not connect to the LMS');
+        this.handleError(new ScormError(CLIENT_COULD_NOT_CONNECT));
       }
 
       return this.lmsConnected;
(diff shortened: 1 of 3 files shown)


js/scorm/wrapper.js
```diff
@@ -465,12 +478,33 @@ define([
...
+        Adapt.once('contentObjectView:ready', this.handleError.bind(this, error));
+        return;
+      }
+
+      const config = Adapt.course.get('_spoor');
+      const errorMessages = Object.assign({}, ScormError.defaultMessages, config && config._errorMessages);
+      const message = Handlebars.compile(errorMessages[error.name])(error.data);
+
+      switch (error.name) {
+        case CLIENT_COULD_NOT_CONNECT:
```
oliverfoster, line 493: @moloko Not sure if you want to handle any other errors this way, or if you want to be able to configure?

oliverfoster:
There are also a series of other logs which may need to be translatable? like: https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/js/scorm/wrapper.js#L75

js/scorm/wrapper.js
moloko: TBH 'could not connect' is the big one... the other one that *might* be useful is [this one](https://github.com/adaptlearning/adapt-contrib-spoor/blob/68eb444bac3926b69da56cb98061b142d3664fe6/js/scorm/wrapper.js#L254)

moloko: ... which I notice now you've already handled!

oliverfoster: 'Course is not connected to the LMS' doesn't display as a notify, only 'Could not connect' is notified atm.

moloko: that's absolutely fine

oliverfoster: I figured I'd do all the ones that end up with a user popup whilst I was there. Not sure how useful that is?

js/scorm/wrapper.js
```diff
@@ -465,12 +478,33 @@ define([
...
+            title: errorMessages['title'],
+            body: message
+          });
+          return;
+      }
+
+      this.logger.error(message);
 
-      if (!this.suppressErrors && (!this.logOutputWin || this.logOutputWin.closed) && confirm(`An error has occured:\n\n${_msg}\n\nPress 'OK' to view debug information to send to technical support.`)) {
+      if (!this.suppressErrors && (!this.logOutputWin || this.logOutputWin.closed) && confirm(`An error has occured:\n\n${message}\n\nPress 'OK' to view debug information to send to technical support.`)) {
```
oliverfoster, line 504: The classic popup pre/post-amble probably needs adding somewhere/how

js/scorm/wrapper.js
moloko: I think it's only right and proper that Adapt, claiming to have multilanguage support, should be able to display error messages in the learner's choice of language!

moloko (approved):

lc-alexanderbenesch (approved):
👍

PR #203, opened 2020-10-15
Allow _commitOnStatusChange setting to be false

tehc-uk:
When publishing a course with Spoor and the setting _commitOnStatusChange set to false, it is not picked up and set during runtime.

It appears as though the if statement around the setting assignment only checks for true values, and as the default value is already true[1], it can never be set to false.

This got us in our older LMS platform which requires this setting to be set to false.

fixes https://github.com/adaptlearning/adapt_framework/issues/2943

[1] https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/js/scorm/wrapper.js#L31

### DIFF
--- a/js/adapt-stateful-session.js
+++ b/js/adapt-stateful-session.js
@@ -54,7 +54,7 @@ define([
       if (settings._suppressErrors) {
         this.scorm.suppressErrors = settings._suppressErrors;
       }
-      if (settings._commitOnStatusChange) {
+      if (settings._commitOnStatusChange === false) {
         this.scorm.commitOnStatusChange = settings._commitOnStatusChange;
       }
       if (settings._commitOnAnyChange) {
--- a/js/adapt-stateful-session.js
+++ b/js/adapt-stateful-session.js
@@ -51,13 +51,13 @@ define([
         this.scorm.showDebugWindow();
       }
       this.scorm.setVersion(settings._scormVersion || '1.2');
-      if (settings._suppressErrors) {
+      if (_.isBoolean(settings._suppressErrors)) {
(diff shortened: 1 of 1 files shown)


js/adapt-stateful-session.js
```diff
@@ -54,7 +54,7 @@ define([
       if (settings._suppressErrors) {
         this.scorm.suppressErrors = settings._suppressErrors;
       }
-      if (settings._commitOnStatusChange) {
+      if (settings._commitOnStatusChange === false) {
```
moloko, line 57:
```suggestion
      if (_.isBoolean(settings._commitOnStatusChange)) {
```
Might be a better solution - just on the off-chance we ever decide to change the default of `scorm.commitOnStatusChange` to `false`.

While you're there, would you mind making the same change for:
* `settings._suppressErrors`
* `settings._commitOnAnyChange`

?

tehc-uk: Yep, looks like a much better solution, thanks. Updated as requested.. :)

moloko (approved):

cahirodoherty-learningpool (approved):
👀

tomgreenfield (approved):
LGTM 👁️.

Not saying this has to be looked at now but the current code is a bit clunky. Everything starting line 54 could become an extensible catch-all:

```js
Object.entries(settings).forEach(([ key, value ]) => {
  key = key.replace(/^_/, '');
  const settingType = typeof this.scorm[key];
  if (settingType === 'undefined' || settingType === 'function') return;
  this.scorm[key] = value;
});
```

PR #208, opened 2020-12-08
reverse logic of return statement in isSupported

moloko:
so that it returns `false` if the error code is 401 and `true` if it's not

fixes https://github.com/adaptlearning/adapt_framework/issues/2991


cahirodoherty-learningpool (approved):
👀

oliverfoster (approved):

tomgreenfield (approved):
:eye:

PR #215, opened 2021-04-26
Hotfix/interaction data underscores

paul-mediakitchen:
The previous update had an issue where the replacing spaces with underscores was not assigned to the variable. This is causing the id to contain spaces which is causing issues on Oyster


oliverfoster:
is this in the wrong place? @paul-mediakitchen

paul-mediakitchen:
wrong branch

PR #218, opened 2021-07-21
issue/3204 Added check for _isTrackable and skip bug warn for missing components

oliverfoster:
fxies https://github.com/adaptlearning/adapt_framework/issues/3204

### Added
* Check for `_isTrackable: false`
* Warn for missing components rather than fail

### DIFF
--- a/js/serializers/ComponentSerializer.js
+++ b/js/serializers/ComponentSerializer.js
@@ -126,7 +126,7 @@ define([
           [model];
         const component = components[index];
         if (!component) {
-          log.warnOnce(`SPOOR could not restore tracking id: ${trackingId}, index: ${index}`);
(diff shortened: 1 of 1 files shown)


js/serializers/ComponentSerializer.js
```diff
@@ -26,7 +27,14 @@ define([
         const components = isContainer ?
           model.findDescendantModels('component') :
           [model];
+
+        let indexOffset = 0;
```
tomgreenfield, line 31: Is the offset needed?

oliverfoster: no, my mistake, thanks.

danielghost (approved):
Resolves the setAttemptObject error.

tomgreenfield (approved):
:eye:

cahirodoherty-learningpool (approved):
👀

PR #222, opened 2021-11-01
#2604 ensure score sent to LMS before status

chris-steele:
Fixes https://github.com/adaptlearning/adapt_framework/issues/2604

See https://github.com/adaptlearning/adapt-contrib-core/pull/30

PR #223, opened 2021-11-23
2808 convert to es6 classes

eleanor-heath:
[#2808](https://github.com/adaptlearning/adapt_framework/issues/2808)

### DIFF
--- a/js/adapt-stateful-session.js
+++ b/js/adapt-stateful-session.js
@@ -34,7 +34,7 @@ define([
       this._shouldStoreAttempts = (tracking && tracking._shouldStoreAttempts) || false;
       // Default should be to record interactions, so only avoid doing that if
       // _shouldRecordInteractions is set to false
-      if (tracking && tracking._shouldRecordInteractions === false) {
+      if (tracking?._shouldRecordInteractions === false) {
         this._shouldRecordInteractions = false;
       }
       const settings = config._advancedSettings;
@@ -108,9 +108,8 @@ define([
           _isAssessmentPassed
         });
       }
-      if (sessionPairs.q) {
-        this._componentSerializer.deserialize(sessionPairs.q);
-      }
+      if (!sessionPairs.q) return;
+      this._componentSerializer.deserialize(sessionPairs.q);
     }
 
     setupEventListeners() {
@@ -142,8 +141,8 @@ define([
       ]);
       const componentStates = this._componentSerializer.serialize(this._shouldStoreResponses, this._shouldStoreAttempts);
       const sessionPairs = {
-        'c': courseState,
-        'q': componentStates
+        c: courseState,
+        q: componentStates
       };
       Adapt.offlineStorage.set(sessionPairs);
       this.printCompletionInformation(sessionPairs);
@@ -193,9 +192,8 @@ define([
       this.removeEventListeners();
       this.setupEventListeners();
       this.saveSessionState();
-      if (config && config._reporting && config._reporting._resetStatusOnLanguageChange === true) {
-        Adapt.offlineStorage.set('status', 'incomplete');
-      }
+      if (config?._reporting?._resetStatusOnLanguageChange === false) return;
+      Adapt.offlineStorage.set('status', 'incomplete');
     }
 
     onVisibilityChange() {
@@ -219,7 +217,7 @@ define([
       const config = Adapt.spoor.config;
       Adapt.course.set('_isAssessmentPassed', stateModel.isPass);
       this.saveSessionState();
-      const shouldSubmitScore = (config && config._tracking && config._tracking._shouldSubmitScore);
+      const shouldSubmitScore = (config?._tracking?._shouldSubmitScore);
       if (!shouldSubmitScore) return;
       const scoreArgs = stateModel.isPercentageBased ?
         [ stateModel.scoreAsPercent, 0, 100 ] :
@@ -235,7 +233,7 @@ define([
       switch (completionData.status) {
         case COMPLETION_STATE.COMPLETED:
         case COMPLETION_STATE.PASSED: {
-          if (!config || !config._reporting || !config._reporting._onTrackingCriteriaMet) {
+          if (!config?._reporting?._onTrackingCriteriaMet) {
             Adapt.log.warn(`No value defined for '_onTrackingCriteriaMet', so defaulting to '${completionStatus}'`);
           } else {
             completionStatus = config._reporting._onTrackingCriteriaMet;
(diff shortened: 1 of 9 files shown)


js/adapt-stateful-session.js
```diff
@@ -1,261 +1,266 @@
...
+    // - get and save a fresh copy of the session state. as the json has been reloaded,
+    //   the blocks completion data will be reset (the user is warned that this will
+    //   happen by the language picker extension)
+    // - check to see if the config requires that the lesson_status be reset to
+    //   'incomplete'
+    const config = Adapt.spoor.config;
+    this.removeEventListeners();
+    this.setupEventListeners();
+    this.saveSessionState();
+    if (config?._reporting?._resetStatusOnLanguageChange === false) return;
```
oliverfoster, line 193: Should probably be.
```suggestion
    if (config?._reporting?._resetStatusOnLanguageChange !== true) return;
```
I think this was a `=== true` because it should only act when explictly set to `true` it'll rarely be set explicitly to `false`

js/adapt-stateful-session.js
```diff
@@ -1,261 +1,266 @@
...
+    Adapt.course.set('_isAssessmentPassed', stateModel.isPass);
+    this.saveSessionState();
+    const shouldSubmitScore = (config?._tracking?._shouldSubmitScore);
+    if (!shouldSubmitScore) return;
+    const scoreArgs = stateModel.isPercentageBased ?
+      [ stateModel.scoreAsPercent, 0, 100 ] :
+      [ stateModel.score, 0, stateModel.maxScore ];
+    Adapt.offlineStorage.set('score', ...scoreArgs);
+  }
+
```
oliverfoster, line 225: This stuff has just been moved out.
https://github.com/adaptlearning/adapt-contrib-spoor/pull/222

js/scorm/wrapper.js
```diff
@@ -1,787 +1,776 @@
...
+    this.logger = Logger.getInstance();
+    this.scorm = pipwerks.SCORM;
+    /**
+     * Prevent the Pipwerks SCORM API wrapper's handling of the exit status
+     */
+    this.scorm.handleExitMode = false;
 
-      return ScormWrapper.instance;
+    this.suppressErrors = false;
+    this.debouncedCommit = setTimeout(this.commit.bind(this), 100);
```
oliverfoster, line 85: This is not equivalent.
https://underscorejs.org/#debounce

In vanilla js a debounce looks something like this:
```js
function debounce (callback, ms) {
  let timeoutHandle = null;
  return function (...args) { // this new function is what's assigned to this.debouncedCommit
    // every time this.debouncedCommit() is called, the previous timeout is cleared and a new one is made
    // such that it won't call `callback` until this.debouncedCommit hasn't been called for at least `ms` time
    if (timeoutHandle) clearTimeout(timeoutHandle);
    timeoutHandle = setTimeout(() => {
      callback.apply(this, args);
    }, ms);
  };
}
```

js/serializers/SCORMSuspendData.js
```diff
@@ -375,7 +375,7 @@ define([
    * @returns {string}
    */
   function binaryToBase64(binary) {
-    binary = _.flatten(binary).join('');
+    binary = Array.prototype.concat.apply([], binary).join('');
```
oliverfoster, line 378: Not equivalent.
https://underscorejs.org/#flatten

js/serializers/SCORMSuspendData.js
```diff
@@ -538,7 +538,8 @@ define([
       this.parent = parent;
       this.name = name;
       this.bitSizes = bitSizes;
-      this.maxValues = _.flatten(this.bitSizes).map(value => (Math.pow(2, value) - 1));
+      this.maxValues = Array.prototype.concat.apply([], this.bitSizes)
```
oliverfoster, line 541: Not equivalent.
https://underscorejs.org/#flatten

js/serializers/SCORMSuspendData.js
```diff
@@ -1062,7 +1063,7 @@ define([
      * @returns {number|boolean|Array}
      */
     valueFromBinary(binary, typeName = null) {
-      binary = _.flatten(binary).join('');
+      binary = Array.prototype.concat.apply([], binary).join('');
```
oliverfoster, line 1066: Not equivalent.
https://underscorejs.org/#flatten

oliverfoster (changes requested):

js/scorm/wrapper.js
```diff
@@ -1,787 +1,776 @@
...
+    this.suppressErrors = false;
+    this.debouncedCommit = _.debounce(this.commit.bind(this), 100);
+    if (window.__debug) {
+      this.showDebugWindow();
     }
 
-    getVersion() {
-      return this.scorm.version;
-    }
+    if (!(window.API?.__offlineAPIWrapper && window?.API_1484_11.__offlineAPIWrapper)) return;
```
cahirodoherty-learningpool, line 90: I think this should be
`if (!(window.API?.__offlineAPIWrapper && window?.API_1484_11?.__offlineAPIWrapper)) return;`

cahirodoherty-learningpool (approved):
👀

oliverfoster (commented):
Looks perfect. Going to give it a quick run throw late tomorrow.

js/serializers/SCORMSuspendData.js
```diff
@@ -940,13 +940,13 @@ define([
     const esTypes = values.map(value => {
       const valueType = findValueTypeFromValue(value, isFloat);
       if (valueType.esType === 'number') {
-        minValue = _.min([value, minValue]);
-        maxValue = _.max([value, maxValue]);
+        minValue = Math.min([value, minValue]);
+        maxValue = Math.max([value, maxValue]);
```
oliverfoster, line 944: `_.min` - https://underscorejs.org/#min
Takes an array or values as the first argument.

`Math.min` - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/min
Takes values as the arguments.

eleanor-heath: could maybe make use of the Math.min shallow copying of the array with `Math.min(...array1)`, so in this case it could be `maxValue = Math.max(...[value, maxValue]);` what do you think?

oliverfoster:
```suggestion
        minValue = Math.min(value, minValue);
        maxValue = Math.max(value, maxValue);
```
This should suffice? Just removing the wrapping array.

oliverfoster (approved):

lc-alexanderbenesch (approved):
👌

PR #228, opened 2022-03-11
issue/1448 Don't report onAssessmentFailed until attempts are used

oliverfoster:
fixes https://github.com/adaptlearning/adapt_framework/issues/1448
### Fixed
* Only report `_onAssessmentFailed` status when no more attempts are possible

Requires https://github.com/adaptlearning/adapt-contrib-assessment/pull/154

### DIFF
--- a/js/adapt-stateful-session.js
+++ b/js/adapt-stateful-session.js
@@ -233,6 +233,7 @@ export default class StatefulSession extends Backbone.Controller {
         break;
       }
       case COMPLETION_STATE.FAILED: {
+        if (completionData.assessment?.canRetry === true) return;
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):

danielghost:
Would it be better for this to be handled in [tracking.js](https://github.com/adaptlearning/adapt-contrib-core/blob/aacbc90cc7eeedaea649270303a695c9fc7f45a0/js/tracking.js), so it is decoupled from the actual plugin and suitable for both SCORM and xAPI etc.?

PR #255, opened 2022-12-08
Fix:  `LZMAFactory` error when Web Workers are not implemented

### LINKED ISSUE #253: LZMA errors if run without Web Workers
If a course is run in an environment without Web Workers, e.g. offline via a mobile app, the course errors on load:
Uncaught TypeError: window.LZMAFactory is not a function

For some reason the LZMA library seems to have been modified to declare `LZMAFactory`, but hasn't implemented this for all possible conditions when the library is initialised.

### Your environment
* since https://github.com/adaptlearning/adapt-contrib-spoor/commit/b7fc02f5e0472b009ef371ad7b6058bee8e0ce62

### Proposed solution
Use the latest libraries files:

- https://github.com/LZMA-JS/LZMA-JS/blob/master/src/lzma-min.js
- https://github.com/LZMA-JS/LZMA-JS/blob/master/src/lzma_worker-min.js

Modify https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/js/serializers/SCORMSuspendData.js#L6 to use `window.LZMA` and prefix all LZMA `compress` and `decompress` references accordingly. Alternatively prefix these methods with `window.LZMA` directly and delete the `LZMAWorker` property.

danielghost:
Fixes #253.


cahirodoherty-learningpool (approved):
👀

oliverfoster (approved):

joe-allen-89 (approved):

PR #248, opened 2022-11-08
Improve use of scorm_test_harness.html (fixes #247)

chris-steele:
#247

### Update

1. make the reset button available so that we may clear tracking data easily
2. allow shift-clicking the reset button to relaunch but preserve tracking data
3. when reset is clicked simulate the user closing and relaunching the course (so no hash/query on the URL)
4. style the reset button in a familiar manner

PR #261, opened 2023-01-11
Fix: Reworked and removed offline_API_wrapper.js (fixes #260)

### LINKED ISSUE #260: Error in offline_API_wrapper.js
MicrosoftTeams-image (2)

offline_API_wrapper.js requires Adapt and is required by Adapt but not compiled into it.

On MacOS, the files can load quickly enough that Adapt isn't finished loading and isn't exposed by requirejs for offline_API_wrapper.js to utilise the Adapt object, resulting in the above error.

oliverfoster:
fixes #260 

### Fix
* Simplified, refactored, separated code
* Normalised HTML files
* Resolved circular loading issues

### Update
* Formalised `_shouldPersistCookieLMSData`, which previously could only be achieved by editing source files.


cahirodoherty-learningpool (approved):
👀

README.md
```diff
@@ -134,6 +134,9 @@ Determines whether the `cmi.completion_status` is set to "completed" if the asse
 #### \_showCookieLmsResetButton (boolean):
 Determines whether a reset button will be available to relaunch the course and optionally clear tracking data (scorm_test_harness.html only). The default is `false`.
 
+#### \_shouldPersistCookieLMSData (boolean):
+Determines whether to persist the cookie data over browser sessions (scorm_test_harness.html only). The default is `false`.
```
chris-steele, line 138: Can we default this to `true` to preserve existing behaviour? Also it was [mentioned ](https://github.com/adaptlearning/adapt-contrib-spoor/pull/248#discussion_r1016496365) scorm_test_harness is sometimes used in production so it would be expected that data is persisted.

js/scorm/cookieLMS.js
```diff
@@ -0,0 +1,187 @@
...
+import Cookies from 'libraries/js-cookie.js';
+
+/** Start the mock API if window.ISCOOKIELMS exists and isn't null */
+export const shouldStart = (Object.prototype.hasOwnProperty.call(window, 'ISCOOKIELMS') && window.ISCOOKIELMS !== null);
+
+/** Store the data in a cookie if window.ISCOOKIELMS is true, otherwise setup the API without storing data. */
+export const isStoringData = (window.ISCOOKIELMS === true);
+
+export function createResetButton() {
+  $('body').append($('<style id="spoor-clear-button">.spoor-reset-button { position:fixed; right:0px; bottom:0px; } </style>'));
```
eleanor-heath, line 11: I think these two parts of the code could be put into variables, e.g.

`const resetButtonStyle = '<style id="spoor-clear-button">.spoor-reset-button { position:fixed; right:0px; bottom:0px; } </style>';`
`const resetButton = '<button class="spoor-reset-button btn-text">Reset</button>'`

and then both of these appended to the body element in one go instead of appending to the body element twice on separate lines of code

js/scorm/cookieLMS.js
```diff
@@ -0,0 +1,187 @@
...
+  const $button = $('<button class="spoor-reset-button btn-text">Reset</button>');
+  $('body').append($button);
+  $button.on('click', e => {
+    if (!e.shiftKey) {
+      Cookies.remove('_spoor');
+      alert('SCORM tracking cookie has been deleted! Tip: shift-click reset to preserve cookie.');
```
eleanor-heath, line 17: should we return here after the alert?

js/scorm/cookieLMS.js
```diff
@@ -0,0 +1,187 @@
...
+    }
+    window.location = window.location.pathname;
+  });
+}
+
+export function configure() {
+  if (!isStoringData) return;
+  const spoorConfig = Adapt.config.get('_spoor');
+  if (spoorConfig?._showCookieLmsResetButton) createResetButton();
```
eleanor-heath, line 26: if (spoorConfig?._showCookieLmsResetButton) return createResetButton();

js/scorm/cookieLMS.js
chris-steele: We want to route the user in both cases (holding shift simulates a re-launch as the tracking data is preserved).

js/scorm/cookieLMS.js
oliverfoster: The persistCookie stuff needs to done regardless of the button, so no return earlier here.

oliverfoster:
ready

eleanor-heath (approved):

chris-steele (approved):

PR #267, opened 2023-02-21
Update: score logic

### LINKED ISSUE #266: format score values accordingly
It is currently possible for a score below 0 and higher than 100 to cause a SCORM error depending on SCORM versions and assessment config - see https://github.com/adaptlearning/adapt-contrib-core/issues/262.

This responsibility to format the score as required by the tracking specification will now need to be handled by each plugin accordingly.

danielghost:
Fixes #266.

### Update
* Amended score logic so the plugin determines the format to send to the LMS
* Negative question scores are now accounted for
* SCORM 1.2 converts score values to a 0-100 percentage at all times
* SCORM 2004 converts `cmi.score.scaled` using the -1-1 range

### Dependency
* https://github.com/adaptlearning/adapt-contrib-core/pull/305

### DIFF
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -214,19 +214,23 @@ class ScormWrapper {
 
   setScore(score, minScore = 0, maxScore = 100) {
     if (this.isSCORM2004()) {
+      // `raw`, `min`, `max` sum absolute values assigned to questions
       this.setValue('cmi.score.raw', score);
       this.setValue('cmi.score.min', minScore);
       this.setValue('cmi.score.max', maxScore);
-
-      const range = maxScore - minScore;
-      const scaledScore = ((score - minScore) / range).toFixed(7);
-      this.setValue('cmi.score.scaled', scaledScore);
-      return;
+      // range split into negative/positive ranges based on score, rather than minScore-maxScore
+      const range = (score < 0) ? Math.abs(minScore) : maxScore;
+      // `scaled` converted to -1-1 range to indicate negative/positive weighting now that negative values can be assigned to questions
+      const scaledScore = score / range;
+      this.setValue('cmi.score.scaled', scaledScore.toFixed(7));
+    } else {
+      // values will always be converted to 0-100 percentage
+      // negative scores are capped to 0 due to SCORM 1.2 limitations
+      const scaledScore = (score < 0) ? 0 : Math.round((score / maxScore) * 100);
+      this.setValue('cmi.core.score.raw', scaledScore);
+      if (this.isSupported('cmi.core.score.min')) this.setValue('cmi.core.score.min', 0);
+      if (this.isSupported('cmi.core.score.max')) this.setValue('cmi.core.score.max', 100);
     }
-    // SCORM 1.2
-    this.setValue('cmi.core.score.raw', score);
-    if (this.isSupported('cmi.core.score.min')) this.setValue('cmi.core.score.min', minScore);
-    if (this.isSupported('cmi.core.score.max')) this.setValue('cmi.core.score.max', maxScore);
   }
 
   getLessonLocation() {


js/scorm/wrapper.js
```diff
@@ -214,19 +214,23 @@ class ScormWrapper {
...
+      // `scaled` converted to -1-1 range to indicate negative/positive weighting now that negative values can be assigned to questions
+      const scaledScore = score / range;
+      this.setValue('cmi.score.scaled', scaledScore.toFixed(7));
+    } else {
+      // values will always be converted to 0-100 percentage
+      // negative scores are capped to 0 due to SCORM 1.2 limitations
+      const scaledScore = (score < 0) ? 0 : Math.round((score / maxScore) * 100);
+      this.setValue('cmi.core.score.raw', scaledScore);
+      if (this.isSupported('cmi.core.score.min')) this.setValue('cmi.core.score.min', 0);
+      if (this.isSupported('cmi.core.score.max')) this.setValue('cmi.core.score.max', 100);
```
oliverfoster, line 232: As mentioned before, I don't think these should always be scaled. Currently absolute positive values are possible.

danielghost: We can still allow these to be absolute values and add a new config setting to handle this if required, but if the score values are not between the 0-100 range, we would need to convert them to percentages anyway (or add a warning and not send to the LMS). Is it preferable to make a consistent rule instead that they will always be percentages? We are already currently making an assumption that the score values are tied to an assessment and not some other form of logic.

oliverfoster: If they've configured to not be percentage and it exceeds the bounds, then it should warn and send anyway. Which is the current behaviour + a new warn.

js/scorm/wrapper.js
```diff
@@ -214,19 +214,23 @@ class ScormWrapper {
 
   setScore(score, minScore = 0, maxScore = 100) {
```
oliverfoster, line 215:
```suggestion
  setScore(score, minScore = 0, maxScore = 100, isPercentageBased = true) {
```

oliverfoster: Plus the other logic as discussed.

danielghost:
Some context to the proposed normalization of scores in this PR, to account for the potential of negative scores. After discussion with @oliverfoster, it was felt that a score of 0 should always be 0. With the standard min-max normalization, it is possible for 0 to be converted to a negative/positive number depending on the score distrubution within the content. The proposed logic ensures this doesn't happen in the following ways:

- SCORM 1.2 caps negative scores at 0
- SCORM 2004 uses a negative or positive range when calculating `cmi.score.scaled`. A negative score will be converted into a negative percentage from the min-0 range. A positive score will be converted into a positive percentage from the 0-max range.

Here is an example of how scores would be converted using the proposed normalization, when using a minimum score of -40 and maximum score of 200:

scores | SCORM 1.2<br>`cmi.core.score.raw` | SCORM 2004<br>`cmi.score.scaled`
:---: | :---: | :---:
-40 | 0 | -1
-20 | 0 | -0.5
0 | 0 | 0
20 | 10 | 0.1
50 | 25 | 0.25
100 | 50 | 0.5
200 | 100 | 1

Using the same example above, this is the conversion using the more typical min-max normalization. For SCORM 2004 examples are shown for both 0..1 and -1..1 normalizations:

scores | SCORM 1.2<br>`cmi.core.score.raw` | SCORM 2004<br>`cmi.score.scaled`<br>0..1 normalization | SCORM 2004<br>`cmi.score.scaled`<br>-1..1 normalization
:---: | :---: | :---: | :---:
-40 | 0 | 0 | -1
-20 | 8.3 | 0.0833333 | -0.8333333
0 | 16.7 | 0.1666667 | -0.6666667
20 | 25 | 0.25 | -0.5
50 | 37.5 | 0.375 | -0.25
100 | 58 | 0.5833333 | 0.1666667
200 | 100 | 1 | 1

Please let us know your thoughts and if you have a preference on the normalizations.

oliverfoster (approved):

js/scorm/wrapper.js
```diff
@@ -212,21 +212,39 @@ class ScormWrapper {
...
+      // `raw`, `min`, `max` sum absolute values assigned to questions
       this.setValue('cmi.score.raw', score);
       this.setValue('cmi.score.min', minScore);
       this.setValue('cmi.score.max', maxScore);
+      // range split into negative/positive ranges (rather than minScore-maxScore) depending on score
+      const range = (score < 0) ? Math.abs(minScore) : maxScore;
+      // `scaled` converted to -1-1 range to indicate negative/positive weighting now that negative values can be assigned to questions
+      const scaledScore = score / range;
+      this.setValue('cmi.score.scaled', scaledScore.toFixed(7));
+    } else {
```
eleanor-heath, line 226: minor: `else {` should go on a new line

js/scorm/wrapper.js
```diff
@@ -212,21 +212,39 @@ class ScormWrapper {
...
+      if (isPercentageBased) {
+        // convert values to 0-100 range
+        // negative scores are capped to 0 due to SCORM 1.2 limitations
+        score = (score < 0) ? 0 : Math.round((score / maxScore) * 100);
+        minScore = 0;
+        maxScore = 100;
+      } else {
```
eleanor-heath, line 233: same here

eleanor-heath (approved):

oliverfoster:
I agree @eleanor-heath, the double nesting is messy.

```js
    if (this.isSCORM2004()) {
      // `raw`, `min`, `max` sum absolute values assigned to questions
      this.setValue('cmi.score.raw', score);
      this.setValue('cmi.score.min', minScore);
      this.setValue('cmi.score.max', maxScore);
      // range split into negative/positive ranges (rather than minScore-maxScore) depending on score
      const range = (score < 0) ? Math.abs(minScore) : maxScore;
      // `scaled` converted to -1-1 range to indicate negative/positive weighting now that negative values can be assigned to questions
      const scaledScore = score / range;
      this.setValue('cmi.score.scaled', scaledScore.toFixed(7));
      return;
    }
    if (isPercentageBased) {
      // convert values to 0-100 range
      // negative scores are capped to 0 due to SCORM 1.2 limitations
      score = (score < 0) ? 0 : Math.round((score / maxScore) * 100);
      minScore = 0;
      maxScore = 100;
    } else {
      const validate = (attribute, value) => {
        const isValid = value >= 0 && score <= 100;
        if (!isValid) this.logger.warn(`${attribute} must be between 0-100.`);
      }
      validate('cmi.core.score.raw', score);
      validate('cmi.core.score.min', minScore);
      validate('cmi.core.score.max', maxScore);
    }
    this.setValue('cmi.core.score.raw', score);
    if (this.isSupported('cmi.core.score.min')) this.setValue('cmi.core.score.min', minScore);
    if (this.isSupported('cmi.core.score.max')) this.setValue('cmi.core.score.max', maxScore);
```

eleanor-heath (approved):

cahirodoherty-learningpool (approved):
👀

PR #270, opened 2023-03-15
New: network connection test

### LINKED ISSUE #268: Loss of Connection Notification feature
### Subject of the feature
Many Learning Management Systems fall short of informing learners of a connection loss which leads to missing or incomplete data for many users. Would like to include a notify message once connection issues occur, which will help users to correct SCORM progress issues as they occur.

### Expected behaviour
A notify pop-up should occur at the point the connection between Adapt and the LMS fails. We cannot fully trust the SCORM API to give factual connection issues, so the proposal would be to test the connection to the server instead.

### Actual behaviour
Currently, little is done within Adapt to help users troubleshoot connection issues once they occur.

danielghost:
Fixes #268 

### New
* Added feature to test the network connection when transferring data to the LMS API.
* Raises an error if the connection is lost.

Further error handling can be utlised with [adapt-contrib-trackingErrors](https://github.com/adaptlearning/adapt-contrib-trackingErrors).

### DIFF
--- a/js/adapt-stateful-session.js
+++ b/js/adapt-stateful-session.js
@@ -51,40 +51,7 @@ export default class StatefulSession extends Backbone.Controller {
       this.scorm.initialize();
       return;
     }
-    if (settings._showDebugWindow) {
-      this.scorm.showDebugWindow();
-    }
-    this.scorm.setVersion(settings._scormVersion || '1.2');
-    if (_.isBoolean(settings._suppressErrors)) {
-      this.scorm.suppressErrors = settings._suppressErrors;
-    }
-    if (_.isBoolean(settings._commitOnStatusChange)) {
-      this.scorm.commitOnStatusChange = settings._commitOnStatusChange;
-    }
-    if (_.isBoolean(settings._commitOnAnyChange)) {
-      this.scorm.commitOnAnyChange = settings._commitOnAnyChange;
-    }
-    if (_.isFinite(settings._timedCommitFrequency)) {
-      this.scorm.timedCommitFrequency = settings._timedCommitFrequency;
-    }
-    if (_.isFinite(settings._maxCommitRetries)) {
-      this.scorm.maxCommitRetries = settings._maxCommitRetries;
-    }
-    if (_.isFinite(settings._commitRetryDelay)) {
-      this.scorm.commitRetryDelay = settings._commitRetryDelay;
-    }
-    if ('_exitStateIfIncomplete' in settings) {
-      this.scorm.exitStateIfIncomplete = settings._exitStateIfIncomplete;
-    }
-    if ('_exitStateIfComplete' in settings) {
-      this.scorm.exitStateIfComplete = settings._exitStateIfComplete;
-    }
-    if (_.isBoolean(settings._setCompletedWhenFailed)) {
-      this.scorm.setCompletedWhenFailed = settings._setCompletedWhenFailed;
-    }
-    const connectionTest = settings._connectionTest;
-    if (connectionTest) Object.assign(this.scorm.connectionTest, connectionTest);
-    this.scorm.initialize();
+    this.scorm.initialize(settings);
   }
 
   restoreSession() {
--- a/js/adapt-stateful-session.js
+++ b/js/adapt-stateful-session.js
@@ -82,6 +82,8 @@ export default class StatefulSession extends Backbone.Controller {
     if (_.isBoolean(settings._setCompletedWhenFailed)) {
       this.scorm.setCompletedWhenFailed = settings._setCompletedWhenFailed;
     }
+    const connectionTest = settings._connectionTest;
+    if (connectionTest) Object.assign(this.scorm.connectionTest, connectionTest);
     this.scorm.initialize();
   }
 
--- a/js/scorm/Connection.js
+++ b/js/scorm/Connection.js
@@ -22,23 +22,20 @@ export default class Connection {
     if (!this._isEnabled || this._isInProgress) return;
     this._isInProgress = true;
 
-    fetch(`connection.json?nocache=${Date.now()}`)
-    .then(response => {
-      (response?.ok) ? this.onConnectionSuccess() : this.onConnectionError();
-    })
-    .catch(error => {
-      this.onConnectionError();
-    });
+    try {
+      const response = await fetch(`connection.json?nocache=${Date.now()}`);
+      if (response?.ok) return this.onConnectionSuccess();
+    } catch (err) {}
+    this.onConnectionError();
   }
 
   reset() {
     this._silentRetryCount = 0;
     this._isSilentDisconnection = false;
 
-    if (this._silentRetryTimeout !== null) {
-      window.clearTimeout(this._silentRetryTimeout);
-      this._silentRetryTimeout = null;
-    }
+    if (this._silentRetryTimeout === null) return;
+    window.clearTimeout(this._silentRetryTimeout);
+    this._silentRetryTimeout = null;
   }
 
   stop() {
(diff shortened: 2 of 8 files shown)


js/scorm/Connection.js
```diff
@@ -0,0 +1,78 @@
...
+    if (!this._isEnabled || this._isInProgress) return;
+    this._isInProgress = true;
+
+    fetch(`connection.json?nocache=${Date.now()}`)
+    .then(response => {
+      (response?.ok) ? this.onConnectionSuccess() : this.onConnectionError();
+    })
+    .catch(error => {
+      this.onConnectionError();
+    });
```
oliverfoster, line 31:
```suggestion
    try {
      const response = await fetch(`connection.json?nocache=${Date.now()}`);
      if (response?.ok) return this.onConnectionSuccess();
    } catch (err) {}
    this.onConnectionError();
```

js/scorm/Connection.js
```diff
@@ -0,0 +1,78 @@
...
+    this._isDisconnected = true;
+
+    if (this._silentRetryCount < this._silentRetryLimit) {
+      this._isSilentDisconnection = true;
+      this._silentRetryCount++;
+      this._silentRetryTimeout = window.setTimeout(this.test.bind(this), this._silentRetryDelay);
+    } else {
+      this.reset();
+      Adapt.trigger('tracking:connectionError', this.test.bind(this));
+    }
```
oliverfoster, line 75:
```suggestion
    if (this._silentRetryCount < this._silentRetryLimit) {
      this._isSilentDisconnection = true;
      this._silentRetryCount++;
      this._silentRetryTimeout = window.setTimeout(this.test.bind(this), this._silentRetryDelay);
      return;
    }
    this.reset();
    Adapt.trigger('tracking:connectionError', this.test.bind(this));
```

js/scorm/wrapper.js
```diff
@@ -511,12 +540,18 @@ class ScormWrapper {
 
     switch (error.name) {
       case CLIENT_COULD_NOT_CONNECT:
-        notify.popup({
-          _isCancellable: false,
-          title: messages.title,
-          body: message
-        });
-        return;
```
oliverfoster, line 519: Was the return intentially remove?

joe-replin:
The connection test is triggering appropriately in my LMS testing. While I have previously seen the data error in a live setting triggering appropriately, I'm not entirely sure how to appropriately test an initialization, data or termination error without triggering the events in the console.

Lastly, I am getting a slight warning in logger.js to change the four instances of 'str: str' to a simple 'str' shorthand.
https://github.com/adaptlearning/adapt-contrib-spoor/blob/a1386cef8bab6e5bae01bd6a44f2ff26d6d670e1/js/scorm/logger.js#L24

joe-replin:
@danielghost, I also tested this by disabling adapt-contrib-trackingErrors. I haven't been given any indication that I do have a connection error even after attempting to adjust my level of logging. Am I to expect a console or alert message without it? 

I understand that some LMSs (e.g. SCORM Cloud) do a good job of notifying users of a network issue via their own popups, so we aren't going to use this feature everywhere. But I suppose a console log would be sufficient for when adapt-contrib-trackingErrors is not attached to the framework.

joe-replin:
I've committed schemas for the connection test feature. I'm not sure if we'll need the legacy schema anymore and this plug-in is currently not compatible with our legacy testing site. I'm also unable to test on a v1 testing site at the moment, but will test these as son as I have an instance available to use.

joe-replin:
@danielghost, one more thing I discovered in testing is that the framework won't build unless "_connectionTest" is part of config. I think you just need to add a simple check to see if it exists wherever the check is made.

js/scorm/Connection.js
oliverfoster: I meant to say: Don't use promises with callbacks when await / async + try/catch is available. Promises and their apis (then/catch) are extremely nesty and incrediblty hard to follow, you'll see eleanor has commented below that it should be further indented with old promise syntax style.

What's the issue with eslint and local errors?

js/scorm/wrapper.js
danielghost: Yes, this was intentional so we can additionally see the logger and debug window, if enabled.

js/scorm/Connection.js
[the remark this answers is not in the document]
danielghost: Fetch syntax has been refactored so comment is no longer relevant.

oliverfoster: await / async is the way of turning promise-based javascript back into that linear, traditional, one line at a time flow, from before promises, instead of handling the flow in callbacks. Much of the way expressions and statements, assignments, errors etc revert to normal, with a few weird exceptions.
```js
[].filter(function predicate (value, index) {

});
```
Filter is a higher-order function, which expects a reference to a non-async predicate function as its first argument. Such that it won't wait if that first argument function references an async function.

```js
[].filter(async function predicate (value, index) {
  let rtnValue = false;
  // do something async
  return rtnValue;
});
```
Filter is expecting the predicate function to return a truthy or a falsey, instead here the predicate function is returning a promise object, as do all async functions. await is the keyword to wait for and extract the value or extract and throw the error from that promise object. Hence why try-catch works with await.
The above code would have to be written using an iteration style that's await/async compatible, like`for`.
```js
const arr = [];
const filteredArr = [];
for (const value of arr) {
  let rtnValue = false;
  // do something async
  if (rtnValue) filteredArr.push(value);
}
```

danielghost:
@joe-replin - please see comments below:

> @danielghost, I also tested this by disabling adapt-contrib-trackingErrors. I haven't been given any indication that I do have a connection error even after attempting to adjust my level of logging. Am I to expect a console or alert message without it?

Good point. Have included this in https://github.com/adaptlearning/adapt-contrib-spoor/pull/270/commits/4dd09c5df1496cb20d89a95e78e42f1b63e87623.

> Lastly, I am getting a slight warning in logger.js to change the four instances of 'str: str' to a simple 'str' shorthand.

Not seeing this, but it shouldn't be linked to the connection test changes, so you may need to investigate this further and raise as a separate issue if it can be replicated.

> @danielghost, one more thing I discovered in testing is that the framework won't build unless "_connectionTest" is part of config. I think you just need to add a simple return check to see if it exists wherever the check is made.

Not seeing this issue myself. If the config is missing it will use the defaults, otherwise it merges them together.

eleanor-heath (approved):

required/connection.json
```diff
@@ -0,0 +1,3 @@
+{
```
oliverfoster, line 1: rename from `connection.json` to `connection.txt` and change content to 1 character, literally, 1, should do it.

joe-replin (approved):
This looks great. :)

eleanor-heath (approved):

js/scorm/wrapper.js
```diff
@@ -116,13 +121,17 @@ class ScormWrapper {
     this.lmsConnected = this.scorm.init();
 
     if (!this.lmsConnected) {
-      this.handleError(new ScormError(CLIENT_COULD_NOT_CONNECT));
+      this.handleInitializeError();
       return this.lmsConnected;
     }
 
+    if (this.connectionTest._isEnabled) {
+      Adapt.on('tracking:connectionError', () => this.handleConnectionError(false));
```
oliverfoster, line 129: This logic create a really weird loop between the `Connection` class and the `handleConnectionError` function and the `tracking:connectionError` event.
It would be preferable to call the `tracking:connectionError` from one place only.

oliverfoster (commented):
See https://github.com/adaptlearning/adapt-contrib-spoor/pull/273 for suggested amends

js/adapt-stateful-session.js
```diff
@@ -82,6 +82,8 @@ export default class StatefulSession extends Backbone.Controller {
     if (_.isBoolean(settings._setCompletedWhenFailed)) {
       this.scorm.setCompletedWhenFailed = settings._setCompletedWhenFailed;
     }
+    const connectionTest = settings._connectionTest;
+    if (connectionTest) Object.assign(this.scorm.connectionTest, connectionTest);
```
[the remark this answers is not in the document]
danielghost: Just added the config to the same file location as all the other overrides, but agree it makes more sense to handle this in the wrapper and connection classes directly.

oliverfoster (approved):

Issue #271, opened 2023-03-28
Course not completed in SCORM - printCompletionInformation - Array out of bound


jonas-jelinski:
### Subject of the issue/enhancement/features
In ${completionString}` the last value of the array becomes always 0.

Reason:

if `markers[trackingId]` is `undefined`

because `trackingId `>= `markers.length`

the code

` markers[trackingId] =
          markers[trackingId] === "-" || markers[trackingId] === "1"
            ? mark
            : "0";`

sets the value `markers[trackingId] ` to 0

### Your environment
* data-adapt-framework-version="5.18.5"
* Firefox 111.0.1 (64-bit)
* Laptop + W10

### Steps to reproduce
Don't know.

### Expected behaviour
Last value in completionString should be 1.

### Actual behaviour
Last value in completionString becomes always 0.

### Screenshots (if you can)

![adapt_debug_spoor_1](https://user-images.githubusercontent.com/67950445/228225595-527d029e-9807-447b-9b69-898523e6ed51.PNG)

![adapt_debug_spoor_2](https://user-images.githubusercontent.com/67950445/228225863-b13378fa-6787-4d5d-98f2-753f1522d2fd.PNG)

oliverfoster:
I don't understand. That function is just for showing the state as a string of 0/1/-, it has nothing to do with completion. Is the course completing on the lms but the console log say's it isn't, or is the course not completing and the console log is also not completing?

https://github.com/adaptlearning/adapt-contrib-spoor/blob/c1b09c1f44776f144f8ff430b2fb80e952a19b17/js/adapt-stateful-session.js#L162-L191


`const max` is the maxiumum`trackingId`, they live at `item[0][0]`
https://github.com/adaptlearning/adapt-contrib-spoor/blob/c1b09c1f44776f144f8ff430b2fb80e952a19b17/js/adapt-stateful-session.js#L177

`const trackingId` reads the same value from each `item[0][0]`
https://github.com/adaptlearning/adapt-contrib-spoor/blob/c1b09c1f44776f144f8ff430b2fb80e952a19b17/js/adapt-stateful-session.js#L180

An array of characteris made using this line, which is one character too short with `new Array(max + 1).join('-').split('')`, it should be `max+2`, one for index to length conversion, and one for the split+join losing a place. Or it could just be changed to `new Array(max + 1).fill('-')`, which does the same thing without the confusion.
[...]

PR #272, opened 2023-03-29
Fix: Correct printCompletionInformation string creation

oliverfoster:
fixes https://github.com/adaptlearning/adapt-contrib-spoor/issues/271

### Fix
* Correct printCompletionInformation string creation, off by one error in array length


cahirodoherty-learningpool (approved):
👀

danielghost (approved):

PR #278, opened 2023-05-31
New: Added _uniqueInteractionIds, testing support for cmi.interactions

### LINKED ISSUE #234: interactions id must be unique
The SCORM 2004 spec is reasonably clear that each interactions record must use a unique identifier. On most LMSes it seems that this is not enforced, but evidentally some do enforce it.

The relevant excerpt from the spec is:

***The cmi.interactions.n.id data model element is a label for the interaction. This label shall be unique at least within the scope of the SCO per interaction record.***

### LINKED ISSUE #277: cmi.interactions.n.id prefix
On https://github.com/adaptlearning/adapt-contrib-spoor/pull/246/files :
image

Each `cmi.interactions.n.id` entry has included an index prefix in order for each interactions entry to be unique. The SCORM spec can be open to interpretation in places, but a necessity for each id to be unique is not clear. 

[1.2 Runtime environment spec, 3-48](https://github.com/adaptlearning/scorm_docs/blob/master/SCORM%201.2/SCORM_1.2_RunTimeEnv.pdf):
1. MicrosoftTeams-image (4)

[2004 v4 runtime environment spec, RTE-4-45](https://github.com/adaptlearning/scorm_docs/blob/master/SCORM%202004/4th%20Edition/SCORM_2004_4ED_v1_1_RTE_20090814.pdf):

1. MicrosoftTeams-image (5)

2. MicrosoftTeams-image (6)

3. MicrosoftTeams-image (8)

4. MicrosoftTeams-image (9)

5. MicrosoftTeams-image (10)

We use the journalling scheme, where every question attempt forms a new entry in the `cmi.interactions` collection.

Whether the identifier `cmi.interactions.n.id` needs to be unique in the SCO and in the collection is perhaps the contention.

Personally, I think it just means that each interaction (i.e. question id) within the content needs to be unique and consistent for all users of that SCO.

If the id is unique for each entry, and doesn't match directly to a question id, it makes analysing and grouping specific interactions/questions more difficult on the server-side.

This has been discussed internally at Kineo, and it is felt that a configuration option can be included to add a unique prefix to `cmi.interactions.n.id`, where required, for any LMS which has interpreted the spec to require each `cmi.interactions.n.id` entry to be unique.

The default should revert back to using the question id for  `cmi.interactions.n.id` only, such that question interaction attempts can be grouped by question id at the server-side. It had previously stood like this for about 8 years, so adding uniqueness could be considered the edge-case worthy of a separate config.

oliverfoster:
fixes #277 

### New
* `config.json:_spoor._advancedSettings._uniqueInteractionIds = false` to allow control over unique `cmi.interactions.n.id` entries
* Switched cookeLMS to store values in nested objects and arrays rather than as a hash of key value pairs (backward compatible), such that it's possible to properly test the `cmi.interactions` property. In the future `cmi.objectives`, `cmi.comments_from_learner` and `cmi.interactions.n.objectives.m` can be easily added to the `scorm_test_harness.html` cookie lms as a result

The cookie lms will look like this now:
image

Instead of this:
image

[SCORM 2004 v4 Spec, RTE-4-5](https://github.com/adaptlearning/scorm_docs/blob/master/SCORM%202004/4th%20Edition/SCORM_2004_4ED_v1_1_RTE_20090814.pdf):
Only support for `cmi.interactions` has been added to the testing environment.
1. image
2. image

### Testing
Running from scorm_test_harness.html, in the console, `window.API.LMSGetValue('cmi.interactions._count')` should return `0` initially and then the number of interactions if any questions are answered.

### DIFF
--- a/js/adapt-stateful-session.js
+++ b/js/adapt-stateful-session.js
@@ -19,6 +19,7 @@ export default class StatefulSession extends Backbone.Controller {
     this._shouldStoreResponses = true;
     this._shouldStoreAttempts = false;
     this._shouldRecordInteractions = true;
+    this._uniqueInteractionIds = false;
     this.beginSession();
   }
 
@@ -51,6 +52,7 @@ export default class StatefulSession extends Backbone.Controller {
       this.scorm.initialize();
       return;
     }
+    this._uniqueInteractionIds = settings._uniqueInteractionIds || false;
     this.scorm.initialize(settings);
   }
 
@@ -177,7 +179,9 @@ export default class StatefulSession extends Backbone.Controller {
     // If responseType doesn't contain any data, assume that the question
     // component hasn't been set up for cmi.interaction tracking
     if (_.isEmpty(responseType)) return;
-    const id = `${this.scorm.getInteractionCount()}-${questionModel.get('_id')}`;
+    const id = this._uniqueInteractionIds
+      ? `${this.scorm.getInteractionCount()}-${questionModel.get('_id')}`
+      : questionModel.get('_id');
     const response = (questionModel.getResponse ? questionModel.getResponse() : questionView.getResponse());
     const result = (questionModel.isCorrect ? questionModel.isCorrect() : questionView.isCorrect());
     const latency = (questionModel.getLatency ? questionModel.getLatency() : questionView.getLatency());
--- a/js/scorm/cookieLMS.js
+++ b/js/scorm/cookieLMS.js
@@ -99,6 +99,13 @@ export function start () {
 
       this.data = Cookies.getJSON('_spoor');
 
+      if (!this.data) {
+        this.data = {};
+        Object.entries(defaults).forEach(([name, value]) => set(this.data, name, value));
+        this.store(true);
+        return false;
+      }
+
       const entries = Object.entries(this.data);
       const isUsingLegacyKeys = (entries[0][0].includes('.'));
       if (isUsingLegacyKeys) {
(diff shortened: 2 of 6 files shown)


js/scorm/cookieLMS.js
```diff
@@ -89,30 +133,42 @@ export function start () {
...
       return 'true';
     },
 
     LMSFinish: function() {
       return 'true';
     },
 
-    LMSGetValue: function(key) {
-      return this.data[key];
+    LMSGetValue: function(path) {
```
cahirodoherty-learningpool, line 150: Can't we just call `this.GetValue(path, value)` if the logic is identical
Likewise for LMSSetValue and this.SetValue()

oliverfoster: Yes. Agreed.

js/adapt-stateful-session.js
```diff
@@ -177,7 +179,9 @@ export default class StatefulSession extends Backbone.Controller {
     // If responseType doesn't contain any data, assume that the question
     // component hasn't been set up for cmi.interaction tracking
     if (_.isEmpty(responseType)) return;
-    const id = `${this.scorm.getInteractionCount()}-${questionModel.get('_id')}`;
+    const id = this._uniqueInteractionIds
+      ? `${this.scorm.getInteractionCount()}-${questionModel.get('_id')}`
+      : questionModel.get('_id');
```
danielghost, line 184: Should all `_uniqueInteractionIds` logic be in https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/js/scorm/wrapper.js instead?

oliverfoster: No, cos it's not a setting for the wrapper, it's a setting for the recordInteractions function in the session.

cahirodoherty-learningpool (approved):
👀

PR #280, opened 2023-06-19
New: Added `cmi.objectives` support (fixes #279).

### LINKED ISSUE #279: Add `cmi.objectives` support
Now that we have the ability to configure multiple scoring sets, it would make sense to add support for `cmi.objectives` to allow scores and completion to be recorded for each scoring set, where supported by an LMS.

Currently the content can only record the overall score and completion, which doesn't offer the granular detail for how the score was achieved in designs using more than one scoring set.

danielghost:
Fixes #279.

### New
* Added `cmi.objectives` support.
  * Objectives for `contentObjects` included.
  * Objectives for `ScoringSets` to be included in the relevant plugins via `OfflineStorage`.

Objective lifecycle as follows:
* Defined and initialised on `adapt:start`, setting the `id`, `description` (SCORM 2004 only) and completion status to "not attempted". This will allow reports to have visibility of all objectives regardless of their current state. In the case of multiple languages, the `description` should be updated accordingly to match the current language.
* Statuses and scores to be updated as necessary, e.g. on visiting/completing a page or assessment.

### Dependencies
 * https://github.com/adaptlearning/adapt-contrib-core/pull/437
 * https://github.com/adaptlearning/adapt-contrib-core/pull/438


js/adapt-stateful-session.js
```diff
@@ -161,7 +165,25 @@ export default class StatefulSession extends Backbone.Controller {
...
   }
 
+  initializeContentObjectives() {
+    Adapt.contentObjects.forEach(model => {
+      if (model.isTypeGroup('course')) return;
+      const id = model.get('_id');
+      const description = model.get('title') || model.get('displayTitle');
+      offlineStorage.set('objectiveDescription', id, description);
+      if (model.get('_isVisited')) return;
+      const completionStatus = 'not attempted';
```
eleanor-heath, line 175: should 'not attempted' be added as an enum to COMPLETION_STATE?

oliverfoster: there's at least one "incomplete" that can also be derived from the enum

js/scorm/wrapper.js
```diff
@@ -690,6 +677,105 @@ class ScormWrapper {
...
+        case 'completed':
+        case 'incomplete':
+        case 'not attempted':
+        case 'not_attempted': // mentioned in SCORM 2004 spec - mapped to 'not attempted'
+        case 'unknown':
+          return true;
+      }
+    } else {
+      switch(status) {
+        case 'passed':
```
oliverfoster, line 754: These could do with being enums rather than strings. It's long winded but at least we get a properly enumerated list of the possibilities.

danielghost: Not sure if all these values are needed. Maybe `COMPLETION_STATE['NOT ATTEMPTED']` but not sure about `COMPLETION_STATE.UNKNOWN`, `COMPLETION_STATE.BROWSED`. We don't want the list to be SCORM specific, although I guess they are still generic terms. Not sure what other context they would be used for.

oliverfoster: You can make a new enum in SPOOR, they don't have to be added to COMPLETION_STATE in the core.

oliverfoster (approved):

eleanor-heath (approved):

joe-allen-89 (approved):
Tested with latest FW (with dependencies) all working as expected.

PR #295, opened 2023-11-07
Fix: Check LMS supports cmi.interactions._count (fixes #294)

### LINKED ISSUE #264: Hardcoded `course` path in script
### Subject of the issue

This should change based upon the `coursedir` parameter at the cli, via `grunt build coursedir=alternative`.

https://github.com/adaptlearning/adapt-contrib-spoor/blob/3955b26fd4a1810cac98adfdc9c804440f718f40/scripts/postcopy.js#L3-L5

### LINKED ISSUE #294: Attempt to getInteractionCount on LMS that doesn't support this SCORM 1.2 feature
I am seeing an issue on an LMS where an error dialogue is displaying after sumitting a question. 

image

Looking at the code I can see it is calling this.scorm.getInteractionCount() inside the onQuestionRecordInteraction function within adapt-stateful-session.js

So it would appear this code does not have any check to see if the LMS supports cmi.interaction._count

I am getting the error every time I answer a question. It is possibly an issue with this particular LMS but it is the only LMS we use that does not have this optional SCORM 1.2 element. So it would be good to hear if anyone else has tested this on an LMS that does not support cmi.interactions._count

https://github.com/adaptlearning/adapt-contrib-spoor/blob/f07a6032a687b6653c4dbcd903797663fa3ec665/js/adapt-stateful-session.js#L205C2-L220C4

paulstevendev:
fixes #294 

### Fix
Add a check that the LMS supports cms.interactions._count to prevent error message appearing to user

### Testing
1. Create a course that includes a question interaction with Record interactions enabled
2. Publish course with the updated Spoor extension and ensure Record interactions is enabled
3. Upload to a LMS that does not have cmi.interactions._count functionality included
4. No error should appear

### DIFF
--- a/js/adapt-stateful-session.js
+++ b/js/adapt-stateful-session.js
@@ -204,6 +204,7 @@ export default class StatefulSession extends Backbone.Controller {
 
   onQuestionRecordInteraction(questionView) {
     if (!this._shouldRecordInteractions) return;
+    if (!this.scorm.isSupported('cmi.interactions._count')) return;
     // View functions are deprecated: getResponseType, getResponse, isCorrect, getLatency
     const questionModel = questionView.model;
     const responseType = (questionModel.getResponseType ? questionModel.getResponseType() : questionView.getResponseType());


oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):
👍

Issue #283, opened 2023-07-07
Issue with SPOOR error Handling message


lokeshece1:
### Undefined spoor configuration error for ScormError

Describe your issue here.

Symptom: Adapt shows endless loop of the the loading screen
Console Error: Seems to be stuck on the cmi.core.lesson_status

### Your environment
* version (Framework 5.31.10) 
* which browser and its version - Chrome
* device(s) + operating system(s) - windows 10

### Steps to reproduce
Check with latest Spoor extension (5.7.0) with different LMS platform

### Expected behaviour
It should first check whether Adapt is enabled then check for the course configuration.
Return early added to wrapper.js if Adapt.course does not exist

[adapt-contrib-spoor/js/scorm/wrapper.js] 
if (!Adapt.course) return;

### Screenshots (if you can)
![image](https://github.com/adaptlearning/adapt-contrib-spoor/assets/35100121/400268bc-56a8-4f2c-873b-8b4afd497cc3)

![image](https://github.com/adaptlearning/adapt-contrib-spoor/assets/35100121/f4222f30-dd1a-4e6a-b472-0de9ef81ebc1)

oliverfoster:
fixed with https://github.com/adaptlearning/adapt-contrib-spoor/pull/301

Issue #299, opened 2023-11-28
Possible issue if course.json does not contain "_spoor" element


paulstevendev:
When testing in the authoring tool using the latest adapt-contrib-spoor I do not see any reference to "_spoor" in the course.json file of the published SCORM zip file.

Therefore I question whether this is going to cause an issue here:

`const config = Adapt.course.get('_spoor');`

https://github.com/adaptlearning/adapt-contrib-spoor/blob/4370780b00d734819159acb866ace308d5ec59f0/js/scorm/wrapper.js#L549C1-L551C74

The code will hang here surely if there is no "_spoor" element present in course.json

paulstevendev:
Possibly related to this issue
https://github.com/adaptlearning/adapt-contrib-spoor/issues/283

paulstevendev:
[course.json](https://github.com/adaptlearning/adapt-contrib-spoor/files/13646739/course.json)
Attached course.json from published course from open source authoring tool using Spoor 5.9.3 and the following
Adapt authoring tool 0.11.2.
Adapt framework v5.33.4
You will see there is no "_spoor" element in course.json

oliverfoster:
Confirmed working with no _spoor in course.json now.

Issue #284, opened 2023-07-19
Update spoor extension to work with jest testing


joe-allen-89:
### Subject of the issue/enhancement/features
- Using import the dependency at the top of the file instead of require
- Return early in spoor's wrapper.js if Adapt.course does not exist

https://github.com/adaptlearning/adapt_framework/pull/3383

PR #301, opened 2023-12-01
Fix: delay error messaging until all course data is ready

### LINKED ISSUE #300: Error messaging lost if course is not yet loaded
If an error occurs before `Adapt.course` has loaded, the error message will be lost. Instead of returning early as per the changes in https://github.com/adaptlearning/adapt-contrib-spoor/commit/cb3d93faecee4fa046088c192b1efd6700d0630d, the error handling should be delayed by adding:
```Javscript
if (!Data.isReady) await Data.whenReady();
```

danielghost:
Fixes #300, #283.

### Fix
* Delay error messaging until all course data is ready.
* Allows other plugins handling errors to listen to error events before they have been fired.

### DIFF
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -517,61 +517,59 @@ class ScormWrapper {
   async handleInitializeError() {
     if (!Data.isReady) await Data.whenReady();
     Adapt.trigger('tracking:initializeError');
-    // defer error to allow other plugins which may be handling errors to execute
-    _.defer(() => this.handleError(new ScormError(CLIENT_COULD_NOT_CONNECT)));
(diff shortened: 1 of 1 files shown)


danielghost:
@joe-allen-89 - please can you check this doesn't impact the fix applied for #284?

eleanor-heath (approved):

paulstevendev:
Does this fix https://github.com/adaptlearning/adapt-contrib-spoor/issues/283 
Issue https://github.com/adaptlearning/adapt-contrib-spoor/issues/283 looks to be the same issue that I raised https://github.com/adaptlearning/adapt-contrib-spoor/issues/299 about.
The issue occurs when no custom error messages have been added so there is no reference to "_spoor" in the course.json and hence there is a script error resulting in the course just hanging on the loading screen.

PR #303, opened 2023-12-05
Fix: Changed logic for 401 error handling

### LINKED ISSUE #302: 401 - Not implemented error handling
Currently the SCORM wrapper checks support for non-mandatory elements via an `LMSGetValue`/`GetValue` before attempting an `LMSSetValue`/`SetValue`. For elements with `_children`, such as `cmi.interactions` and `cmi.objectives`, it assumes that if the `_count` is supported, all elements of that data model are supported. Whilst this is generally true, there are edge cases where an LMS doesn't support all child elements, causing Adapt to raise an error.

For future reference the following occurs for Adobe Learning Manager:

**SCORM 1.2**
- `cmi.student_preference.language` is unsupported by the LMS, but incorrectly returns a 201 invalid element error rather than the 401 unsupported error. The error codes are inconsistently implemented with an `LMSGetValue` call to this element incorrectly returning 201, but an `LMSSetValue` correctly returning 401. `cmi.student_preference._children` correctly returns an empty string and a 401 error.
- `cmi.objectives._children` returns “id,score,status” as the supported elements, but `score` and `status` both return 401 errors when you attempt to use them.

**SCORM 2004**
- `cmi.learner_preference` and `cmi.objectives` have not been implemented, yet are mandatory elements. Both incorrectly return “undefined” and a 201 error when calling `_children`.

Should we simplify this approach and remove conditional checks completely and instead handle 401 errors silently? This removes the potential for implementation issues such as #294, with the knowledge that a call to any valid unsupported elements will fail gracefully in the same way they are when `isSupported` calls are made. If we wish to maintain conditional checks to reduce the number of 401 handled errors, the recommended approach would be to check which `_children` are returned first - see #241.

danielghost:
Fixes #302.

### Fix
* Changed logic for 401 error handling so calls to unsupported data elements will fail silently.

### DIFF
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -536,21 +536,21 @@ class ScormWrapper {
 
   async handleError(error) {
     if (!Data.isReady) await Data.whenReady();
-    if ('value' in error.data) {
-      // because some browsers (e.g. Firefox) don't like displaying very long strings in the window.confirm dialog
-      if (error.data.value.length && error.data.value.length > 80) error.data.value = error.data.value.slice(0, 80) + '...';
-      // if the value being set is an empty string, ensure it displays in the error as ''
-      if (error.data.value === '') error.data.value = '\'\'';
-    }
-    const config = Adapt.course.get('_spoor');
-    const messages = Object.assign({}, ScormError.defaultMessages, config && config._messages);
-    const message = Handlebars.compile(messages[error.name])(error.data);
-    this.logger.error(message);
-    if (this.isUnsupportedErrorCode(error.data.errorCode)) return;
-    switch (error.name) {
-      case CLIENT_COULD_NOT_CONNECT:
-        // defer error to allow other plugins which may be handling errors to execute first
-        _.defer(() => {
+    // defer error to allow other plugins which may be handling errors to execute first
+    _.defer(() => {
+      if ('value' in error.data) {
+        // because some browsers (e.g. Firefox) don't like displaying very long strings in the window.confirm dialog
+        if (error.data.value.length && error.data.value.length > 80) error.data.value = error.data.value.slice(0, 80) + '...';
+        // if the value being set is an empty string, ensure it displays in the error as ''
+        if (error.data.value === '') error.data.value = '\'\'';
+      }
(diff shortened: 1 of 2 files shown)


oliverfoster:
Do the `if (!this.scorm.isSupported('cmi.interactions._count')) return;` calls work in some situations currently?

danielghost:
> Do the `if (!this.scorm.isSupported('xxx')) return;` calls work in some situations currently?
> 
> If so, it might be wise to leave them in there rather than making a known 404 call.

They still work, but see https://github.com/adaptlearning/adapt-contrib-spoor/issues/302#issuecomment-1842909533 to help decide if these should remain.

js/scorm/wrapper.js
```diff
@@ -540,19 +536,21 @@ class ScormWrapper {
...
-        if (error.data.value.length && error.data.value.length > 80) error.data.value = error.data.value.slice(0, 80) + '...';
-        // if the value being set is an empty string, ensure it displays in the error as ''
-        if (error.data.value === '') error.data.value = '\'\'';
-      }
-      const config = Adapt.course.get('_spoor');
-      const messages = Object.assign({}, ScormError.defaultMessages, config && config._messages);
-      const message = Handlebars.compile(messages[error.name])(error.data);
-      switch (error.name) {
-        case CLIENT_COULD_NOT_CONNECT:
+    if ('value' in error.data) {
```
oliverfoster, line 539: Why has the defer moved?

danielghost: I think it was because the defer was only added to prevent multiple notifications from being opened at once, e.g. if [adapt-trackingErrors](https://github.com/adaptlearning/adapt-contrib-trackingErrors) is enabled, so it doesn't need to delay the other parts of the function. Happy to revert if necessary as I can't remember another reason why I changed it and can't replicate any potential issues as a result of it remaining on line 544 instead.

oliverfoster: Yes, please revert unless it's related to the issue.

oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):
👍

PR #306, opened 2023-12-11
Fix: Amended logic to check for supported `_children` elements before attempting to set values

### LINKED ISSUE #304: Unsupported logic
In addition to the changes made for #302, the unsupported logic should change to check which child elements are supported, as per the spec:
 > datamodel.element._children
The _children keyword is used to determine all of the elements in a group or category that are supported by the LMS.

> LMSGetValue(datamodel.group._children)
The return value is a comma-separated list of all of the element names in the specified group or category that are supported by the LMS. If an element has no children, but is supported, an empty string ("") is returned. An empty string ("") is also returned if an element is not supported. A subsequent request for last error [LMSGetLastError()] can determine if the element is not supported. The error “401 Not implemented error” indicates the element is not supported.

The current method of just checking `_count` could lead to errors being raised where not all the child elements are supported.

danielghost:
Fixes #304.

### Fix
* Amended logic to check for supported `_children` elements before attempting to set values.

### DIFF
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -636,7 +636,7 @@ class ScormWrapper {
     this.setValue(`${cmiPrefix}.type`, type);
     this.setValue(`${cmiPrefix}.learner_response`, response);
     this.setValue(`${cmiPrefix}.result`, correct ? 'correct' : 'incorrect');
-    if (latency != null) this.setValue(`${cmiPrefix}.latency`, this.convertToSCORM2004Time(latency));
+    if (latency !== null && latency !== undefined) this.setValue(`${cmiPrefix}.latency`, this.convertToSCORM2004Time(latency));
     this.setValue(`${cmiPrefix}.timestamp`, this.getISO8601Timestamp());
   }
 
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -621,8 +621,10 @@ class ScormWrapper {
       : this.logUnsupported(`${cmiPrefix}.student_response`);
     children.includes('result') ? this.setValue(`${cmiPrefix}.result`, correct ? 'correct' : 'wrong')
       : this.logUnsupported(`${cmiPrefix}.result`);
-    children.includes('latency') && latency != null ? this.setValue(`${cmiPrefix}.latency`, this.convertToSCORM12Time(latency))
+    if (latency != null) {
+      children.includes('latency') ? this.setValue(`${cmiPrefix}.latency`, this.convertToSCORM12Time(latency))
       : this.logUnsupported(`${cmiPrefix}.latency`);
+    }
     children.includes('time') ? this.setValue(`${cmiPrefix}.time`, this.getCMITime())
       : this.logUnsupported(`${cmiPrefix}.time`);
   }
@@ -645,7 +647,8 @@ class ScormWrapper {
       response = response.replace(/#/g, ',');
       response = this.checkResponse(response, 'choice');
     }
-    this.isSCORM2004() ? this.recordInteractionScorm2004(...arguments) : this.recordInteractionScorm12(...arguments);
+    const scormRecordInteraction = this.isSCORM2004() ? this.recordInteractionScorm2004 : this.recordInteractionScorm12;
+    scormRecordInteraction.call(this, id, response, correct, latency, type);
   }
 
   recordInteractionMatching(id, response, correct, latency, type) {
@@ -655,7 +658,8 @@ class ScormWrapper {
     } else {
       response = this.checkResponse(response, 'matching');
     }
-    this.isSCORM2004() ? this.recordInteractionScorm2004(...arguments) : this.recordInteractionScorm12(...arguments);
+    const scormRecordInteraction = this.isSCORM2004() ? this.recordInteractionScorm2004 : this.recordInteractionScorm12;
+    scormRecordInteraction.call(this, id, response, correct, latency, type);
   }
 
   recordInteractionFillIn(id, response, correct, latency, type) {
@@ -665,7 +669,8 @@ class ScormWrapper {
       response = response.substr(0, maxLength);
       this.logger.warn(`ScormWrapper::recordInteractionFillIn: response data for ${id} is longer than the maximum allowed length of ${maxLength} characters; data will be truncated to avoid an error.`);
     }
-    this.isSCORM2004() ? this.recordInteractionScorm2004(...arguments) : this.recordInteractionScorm12(...arguments);
+    const scormRecordInteraction = this.isSCORM2004() ? this.recordInteractionScorm2004 : this.recordInteractionScorm12;
+    scormRecordInteraction.call(this, id, response, correct, latency, type);
   }
 
   getObjectiveCount() {
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -273,12 +273,7 @@ class ScormWrapper {
       this.setValue('cmi.learner_preference.language', lang);
       return;
     }
-    const children = this.getValue('cmi.student_preference._children');
-    if (this.isUnsupportedLastError() || !children.includes('language')) {
-      this.logUnsupported('cmi.student_preference.language');
-      return;
-    }
-    this.setValue('cmi.student_preference.language', lang);
+    if (this.isChildSupported('cmi.student_preference.language')) this.setValue('cmi.student_preference.language', lang);
   }
 
   commit() {
@@ -470,7 +465,31 @@ class ScormWrapper {
       return false;
     }
     this.scorm.get(property);
-    return !this.isUnsupportedLastError();
+    const isSupported = !this.isUnsupportedLastError();
+    if (!isSupported) this.logUnsupported(property);
+    return isSupported;
+  }
+
+  isChildSupported(property) {
+    if (property.includes('_children')) return this.isSupported(property);
+    this.logger.debug(`ScormWrapper::isChildSupported: _property=${property}`);
+    if (this.finishCalled) {
+      this.logger.debug('ScormWrapper::isChildSupported: ignoring request as \'finish\' has been called');
+      return;
+    }
+    if (!this.lmsConnected) {
+      this.handleConnectionError();
+      return false;
+    }
+    const paths = property.split('.');
+    const element = paths.pop();
+    // remove last path if contains indexes
+    if (/^\d+$|^n$/.test(paths[paths.length - 1])) paths.pop();
+    const parentPath = paths.join('.');
+    const children = this.scorm.get(`${parentPath}._children`);
+    const isSupported = !this.isUnsupportedLastError() && children.includes(element);
+    if (!isSupported) this.logUnsupported(property);
+    return isSupported;
   }
 
   isUnsupportedErrorCode(code) {
@@ -482,7 +501,7 @@ class ScormWrapper {
   }
 
   logUnsupported(property) {
-    property = property.replace(/[0-9]+/g, 'n');
+    property = property.replace(/\d+/g, 'n');
     this.logger.info(`ScormWrapper::${property} not supported by this LMS...`);
   }
 
@@ -593,12 +612,8 @@ class ScormWrapper {
       validate(`${cmiPrefix}.score.max`, maxScore);
     }
     this.setValue(`${cmiPrefix}.score.raw`, score);
-    const children = this.getValue(`${cmiPrefix}.score._children`);
-    if (this.isUnsupportedLastError()) return;
-    children.includes('min') ? this.setValue(`${cmiPrefix}.score.min`, minScore)
-      : this.logUnsupported(`${cmiPrefix}.score.min`);
-    children.includes('max') ? this.setValue(`${cmiPrefix}.score.max`, maxScore)
-      : this.logUnsupported(`${cmiPrefix}.score.max`);
+    if (this.isChildSupported(`${cmiPrefix}.score.min`)) this.setValue(`${cmiPrefix}.score.min`, minScore);
+    if (this.isChildSupported(`${cmiPrefix}.score.max`)) this.setValue(`${cmiPrefix}.score.max`, maxScore);
   }
 
   getInteractionCount() {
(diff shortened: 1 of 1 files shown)


js/scorm/wrapper.js
```diff
@@ -605,77 +605,60 @@ class ScormWrapper {
...
-    this.setValue(`${cmiPrefix}.type`, type);
-    this.setValue(`${cmiPrefix}.student_response`, response);
-    this.setValue(`${cmiPrefix}.result`, correct ? 'correct' : 'wrong');
-    if (latency !== null && latency !== undefined) this.setValue(`${cmiPrefix}.latency`, this.convertToSCORM12Time(latency));
-    this.setValue(`${cmiPrefix}.time`, this.getCMITime());
+    children.includes('type') ? this.setValue(`${cmiPrefix}.type`, type) : this.logUnsupported(`${cmiPrefix}.type`);
+    children.includes('student_response') ? this.setValue(`${cmiPrefix}.student_response`, response) : this.logUnsupported(`${cmiPrefix}.student_response`);
+    children.includes('result') ? this.setValue(`${cmiPrefix}.result`, correct ? 'correct' : 'wrong') : this.logUnsupported(`${cmiPrefix}.result`);
+    children.includes('latency') && latency != null ? this.setValue(`${cmiPrefix}.latency`, this.convertToSCORM12Time(latency)) : this.logUnsupported(`${cmiPrefix}.latency`);
+    children.includes('time') ? this.setValue(`${cmiPrefix}.time`, this.getCMITime()) : this.logUnsupported(`${cmiPrefix}.time`);
```
danielghost, line 620: Would the following be preferable for readability so the ternary confidtion isn't all on one line?
```suggestion
    children.includes('type') ? this.setValue(`${cmiPrefix}.type`, type)
      : this.logUnsupported(`${cmiPrefix}.type`);
    children.includes('student_response') ? this.setValue(`${cmiPrefix}.student_response`, response)
      : this.logUnsupported(`${cmiPrefix}.student_response`);
    children.includes('result') ? this.setValue(`${cmiPrefix}.result`, correct ? 'correct' : 'wrong')
      : this.logUnsupported(`${cmiPrefix}.result`);
    children.includes('latency') && latency != null ? this.setValue(`${cmiPrefix}.latency`, this.convertToSCORM12Time(latency))
      : this.logUnsupported(`${cmiPrefix}.latency`);
    children.includes('time') ? this.setValue(`${cmiPrefix}.time`, this.getCMITime())
      : this.logUnsupported(`${cmiPrefix}.time`);
```

js/scorm/wrapper.js
```diff
@@ -364,28 +368,19 @@ class ScormWrapper {
...
   recordInteraction(id, response, correct, latency, type) {
-    if (!this.isSupported('cmi.interactions._count')) {
-      this.logger.info('ScormWrapper::recordInteraction: cmi.interactions are not supported by this LMS...');
-      return;
-    }
-
     switch (type) {
       case 'choice':
-        this.recordInteractionMultipleChoice.apply(this, arguments);
+        this.recordInteractionMultipleChoice(...arguments);
```
oliverfoster, line 373: I don't think we could spread the arguments object because it isn't iterable.

https://caniuse.com/mdn-javascript_functions_arguments_--iterator
https://stackoverflow.com/questions/34614653/is-the-arguments-object-supposed-to-be-an-iterable-in-es6

Apparently one can now. Cools.

oliverfoster:
I don't understand what all the other code changes in this pr are for, this would otherwise be something very simple. Can you please roll back all of the other code changes which aren't covered by the main issue and which complicate the merge process and do a pr for only the parts required?

js/scorm/wrapper.js
```diff
@@ -605,76 +607,69 @@ class ScormWrapper {
...
     id = id.trim();
-
     const cmiPrefix = `cmi.interactions.${this.getInteractionCount()}`;
-
     this.setValue(`${cmiPrefix}.id`, id);
     this.setValue(`${cmiPrefix}.type`, type);
     this.setValue(`${cmiPrefix}.learner_response`, response);
     this.setValue(`${cmiPrefix}.result`, correct ? 'correct' : 'incorrect');
-    if (latency !== null && latency !== undefined) this.setValue(`${cmiPrefix}.latency`, this.convertToSCORM2004Time(latency));
+    if (latency != null) this.setValue(`${cmiPrefix}.latency`, this.convertToSCORM2004Time(latency));
```
oliverfoster, line 639: Never use a loose equality operator please. Always the triple equals variety. The `!= null` for undefined and null detection is by no means obvious and isn't well used anywhere else in the code.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null#difference_between_null_and_undefined

oliverfoster (changes requested):

js/scorm/wrapper.js
danielghost: Can change back, but thought this should be clear as that is how the loose equality operator works, as shown in the link provided.

danielghost:
> I don't understand what all the other code changes in this pr are for, this would otherwise be something very simple. Can you please roll back all of the other code changes which aren't covered by the main issue and which complicate the review and merge process and do a pr for only the parts required?

As explained in the description, this also includes the changes from #303, as the new methods added for that PR are required. Once that has been approved and the master branch merged into this PR, it will only include the changes specific to this issue. There are also a few small unrelated code refactors (like the spread arguments) in the methods amended for this issue, which aren't necessary for it to work, but made sense to apply whilst those methods were being updated. These small refactors are regularly included in other PRs. It doesn't include refactoring other methods which weren't already being modified or linked to those which have been.

oliverfoster:
Submitting PRs like this makes reviewing quite difficult. You've effectively shifted the burden onto all of the reviewers to understand what you're doing by changing the simple, usual process with a comment and you've put something that's not ready into the review process. 

It's better if you do clean prs or move this out of needs reviewing until it's ready.

danielghost:
Fair enough, will delay assigning it for review for cases like this in the future.

js/scorm/wrapper.js
```diff
@@ -592,8 +593,12 @@ class ScormWrapper {
...
     }
     this.setValue(`${cmiPrefix}.score.raw`, score);
-    if (this.isSupported(`${cmiPrefix}.score.min`)) this.setValue(`${cmiPrefix}.score.min`, minScore);
-    if (this.isSupported(`${cmiPrefix}.score.max`)) this.setValue(`${cmiPrefix}.score.max`, maxScore);
+    const children = this.getValue(`${cmiPrefix}.score._children`);
+    if (this.isUnsupportedLastError()) return;
+    children.includes('min') ? this.setValue(`${cmiPrefix}.score.min`, minScore)
+      : this.logUnsupported(`${cmiPrefix}.score.min`);
+    children.includes('max') ? this.setValue(`${cmiPrefix}.score.max`, maxScore)
+      : this.logUnsupported(`${cmiPrefix}.score.max`);
```
oliverfoster, line 601: We tend not to use ternaries for conditional statement execution but instead as expressions for generating a value from conditions.

There is a lot of repeating behaviour here which you're trying to reduce with ternaries.

Perhaps add `isChildSupported`?:

```js
if (this.isChildSupported(`${cmiPrefix}.score.min`)) this.setValue(`${cmiPrefix}.score.min`, minScore);

isChildSupported(propertyPath) {
  const propertyPathParts = propertyPath.split('.');
  const childPropertyName = propertyPathParts.pop();
  const parentPropertyPath = propertyPathParts.join('.');
  const children = this.getValue(`${parentPropertyPath}._children`);
  if (this.isUnsupportedLastError()) {
    this.logUnsupported(propertyPath);
    return false;
  }
  return children.includes(childPropertyName);
}
```

oliverfoster (changes requested):

js/scorm/wrapper.js
oliverfoster: > this could result in lots of additional API calls.

With SCORM the API call will only look up sub properties from a nested state object, it will not be a resource hungry process, caching the children variable will not give any noticeable performance improvement, optimisation through variable caching is completely unnecessary here and is causing obtuse code. We're not talking about a slow API, such as one that will cross javascript contexts across Window boundaries or perform HTTP requests, read files from disk etc, this is just traversing local javascript object properties with a few checks. SCORM's getValue and setValue are almost identical to Backbone's model set and get, except that they allow sub properties through dot name paths and have a few additional checks.

[Moodle scorm 1.2 API implementation](https://github.com/moodle/moodle/blob/9587029a4609b3222a3564207aada02716d3332d/mod/scorm/datamodels/scorm_12.js#L242-L406)

danielghost: I understand how the API works and that the SCO isn't making HTTP requests, but I wasn't clear how little the performance would be impacted by making multiple calls compared to one. I thought calling the API `LMSGetValue`/`GetValue` to lookup the models multiple times for the same method would be inefficient. Can we be sure that all LMSs implement the API calls efficiently?

I was also mindful that the SCORM log becomes noisier as a result, making it more obtuse when being used to analyse diagnostics from SCORM Test Packages.

Ultimately, if we are happy to lean on the API more to reduce code, without impacting performance, I can adjust accordingly.

oliverfoster: It's always a cached state object. Sometimes it's a nested hierarchy, sometimes a flat name value hash like on Moodle / Totara. The API is synchronous for get and set and commit and finish are used to push the states to the lms. 

The scorm logging provided by the wrapper in the function `this.getValue`, for a function like `this.isChildSupported`, can be circumvented by using `this.scorm.get` directly. As it is in `this.isSupported` currently. Checking for support in a new function wouldn't be logged as it isn't logged currently.

The performance impact of the optimization in this code would always only be client side and with a high degree of certainty, be always completely negligible. We set suspendData with greater frequency than we'll ever set interactions.

danielghost: `isSupported` does log calls - https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/js/scorm/wrapper.js#L471, but I get your point that it directly targets the pipwerks library, so doesn't add to the log for the get value requests.

Good point regarding `suspend_data` and `lesson_location`/`location` requests which are constantly firing without issue.

oliverfoster:
What it your intent to remove the support check from the generic recordInteraction, covering both 1.2 and 2004, and to move it into the 1.2 function only?

js/scorm/wrapper.js
danielghost: Updated support logic to include unsupported logging in the methods checking support:
https://github.com/adaptlearning/adapt-contrib-spoor/pull/306/commits/0c6836b508df040989a4e693191684137a1970f0

Adding supported checks into `setValue` and returning early if not supported would have been nice and only needed for SCORM 1.2, but would ideally include some form of conditional checking against acceptable property values to ensure we aren't triggering 201 LMS errors for values which don't have `_children`. I didn't really want to include hardcoded references to capture the following models which allow write access to their children:
- cmi.core.score
- cmi.objectives
- cmi.objectives.n.score
- cmi.student_preference
- cmi.interactions

oliverfoster:
> > Was it your intent to remove the support check from the generic recordInteraction, covering both 1.2 and 2004, and to move it into the 1.2 function only?
> 
> Yes, it is a mandatory requirement to be supported in SCORM 2004. This is the case for all data models in SCORM 2004, unlike 1.2.

Could you put it back please. Removing existing checks is out of scope of this issue.

oliverfoster (approved):

js/scorm/wrapper.js
```diff
@@ -273,8 +273,7 @@ class ScormWrapper {
       this.setValue('cmi.learner_preference.language', lang);
       return;
     }
-    if (!this.isSupported('cmi.student_preference.language')) return;
-    this.setValue('cmi.student_preference.language', lang);
+    if (this.isChildSupported('cmi.student_preference.language')) this.setValue('cmi.student_preference.language', lang);
```
cahirodoherty-learningpool, line 276:
```suggestion
    if (this.isChildSupported('cmi.student_preference.language')) {
      this.setValue('cmi.student_preference.language', lang);
    }
```
Minor: I think braces here would make this a little easier to read

js/scorm/wrapper.js
```diff
@@ -478,13 +466,46 @@ class ScormWrapper {
...
+    const paths = property.split('.');
+    const element = paths.pop();
+    // remove last path if contains indexes
+    if (/^\d+$|^n$/.test(paths[paths.length - 1])) paths.pop();
+    const parentPath = paths.join('.');
+    const children = this.scorm.get(`${parentPath}._children`);
+    const isSupported = !this.isUnsupportedLastError() && children.includes(element);
+    if (!isSupported) this.logUnsupported(property);
+    return isSupported;
   }
```
cahirodoherty-learningpool, line 495: Could we add a helper function?

```suggestion
  setValueIfSupported(property, value) {
    if (!this.isChildSupported(property)) {
      this.logUnsupported(property);
      return;
    }
    
      this.setValue(property, value);
  }
```

danielghost: We could, but support for child elements are handled differently to others, so you would need awareness of the children models to ensure the correct supported checks are made before attempting to set the value. Please see https://github.com/adaptlearning/adapt-contrib-spoor/pull/306#discussion_r1450448779.

These changes could be made so you only need to call `setValue` and the wrapper takes care of it for you regardless of the data models being used, but it becomes messier as you have to create those hardcoded references to accepted elements etc., or we keep it simple and expect some knowledge of SCORM and what methods to use, as is the case currently.

oliverfoster: There are 7 instances of `if (this.isChildSupported(path)) this.setValue(path, value);` that could be just `this.setValueIfChildSupported(path, value);`
```js
  setValueIfChildSupported(property, value) {
    if (!this.isChildSupported(property)) return;
    this.setValue(property, value);
  }
  ```

Comment (danielghost) [reply js/scorm/wrapper.js:495]: This is another example of a change that doesn't align with what was already in the file, but can be included. It wasn't a concern previously for the 3 examples where `isSupported` was being checked before setting a value. If going this route, I assume we should also include a `setValueIfSupported` for data models which don't have `_children`, i.e. should `cmi.comments` be used by an author?

My main hesitation for not including it previously was that I didn't want a method to look like you could throw any property at it and it will handle the call without issue. Someone without full understanding of SCORM could potentially look at the API and attempt to use that method outside the wrapper for properties without `_children`, or that don't exist, not realising that LMS errors will be raised. Not that it is a big deal to raise an LMS error, as previously I wanted to remove all checks to supported elements and just handle the LMS errors accordingly (https://github.com/adaptlearning/adapt-contrib-spoor/issues/302).

Maybe this isn't a concern and we don't need to account for something being used incorrectly outside of the plugin. It could be a simple case of changing the API to denote that not all methods are intended to be public and we use private/protected methods such as this to simplify calls inside the wrapper, where we know data models are being used correctly with the appropriate method?

Comment (oliverfoster) [reply js/scorm/wrapper.js:495]: I think that doing what Cahir suggested, condensing the repeated behaviour into a single function call:
```js
  setValueIfSupported(property, value) {
    if (!this.isChildSupported(property)) return;
    this.setValue(property, value);
  }
```
But changing its name to `setValueIfChildSupported` is sufficiently clear to cover all of the raised concerns. If `isChildSupported` is clear enough, which I think it is, assuming some knowledge of SCORM and the relatively isolated nature of this API, then `setValueIfChildSupported` is clear enough.

It is also possible to add jsdoc comments to a function definitions with clarifying descriptions. I don't think that it's necessary in this case however.

I don't think we need to be overly concerned about developers making illegal calls to `setValue` or `isChildSupported` or `setValueIfChildSupported`.

oliverfoster (approved):

joe-allen-89 (approved):
👍

cahirodoherty-learningpool (approved):
👀

PR #308, opened 2024-01-24
Fix: error logging order to match call execution (fixes #307).

### LINKED ISSUE #307: Error logging order
The logged SCORM calls are out of sync when adding errors, due to the deferred handling. The logs would be easier to analyse if the error entries followed the associated get/set calls rather than appearing at the end of the stack.

The reverted change of https://github.com/adaptlearning/adapt-contrib-spoor/pull/303/commits/3f68ac6663df4c80d50477394d274f6d831c82e8 (previously reverted to be handled in a separate PR) should be implemented accordingly so only the notification is deferred, keeping the logs in sync with the execution order.

danielghost:
Fixes #307.

### Fix
* Error logging order to match execution order of SCORM calls.

### DIFF
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -562,21 +562,21 @@ class ScormWrapper {
 
   async handleError(error) {
     if (!Data.isReady) await Data.whenReady();
-    // defer error to allow other plugins which may be handling errors to execute first
-    _.defer(() => {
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

lemmyadams (approved):

oliverfoster (approved):

PR #298, opened 2023-11-14
fix: Adding a pagehide listener to cover bases when unload becomes deprecated

cahirodoherty-learningpool:
[//]: # (Please title your PR according to eslint commit conventions)
[//]: # (See https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-eslint#eslint-convention for details)

[//]: # (Link the PR to the original issue)
Partly fixes [FW3481](https://github.com/adaptlearning/adapt_framework/issues/3481)

[//]: # (Delete Fix, Update, New and/or Breaking sections as appropriate)
### Fix
* Added a 'pagehide' listener to backup the soon-to-be-deprecated 'unload' event

PR #315, opened 2024-04-04
Fix: Add pagehide event (Fixes #314)

### LINKED ISSUE #314: unload event will soon be deprecated in Google Chrome - Add workaround to version of Spoor that is compatible with v4 framework
The unload event will be deprecated in Google Chrome by the end of Q3 2024. This has already been raised (https://github.com/adaptlearning/adapt_framework/issues/3481) and fixes were made to the latest versions of relevant plugins.

We still have a number of courses that use the v4 framework, and so we need the same fix to be applied to the last version of Spoor to support v4 (3.3.2).

### Your environment
* v4 Framework

canstudios-nicolaw:
Fixes https://github.com/adaptlearning/adapt-contrib-spoor/issues/314

### Fix
Added the pagehide event next to the unload event with the same behaviour so that end-users won't notice when Google Chrome stops supporting unload. This is the same fix as was applied in the latest version of Spoor https://github.com/adaptlearning/adapt-contrib-spoor/pull/298/files.

### Testing
To test this I tried removing the unload events entirely and switching to pagehide events. I created and published a course and I was able to see that my progress was saved if I closed the window part-way through.

### Note
I'm not sure if there is a process for updating an older version of a plugin like this. I created a release-3.3 branch from the v3.3.2 tag. I have made a pull request to add this fix to the release-3.3 branch and I propose we tag v3.3.3 from the release-3.3 branch. I'm open to other suggestions if this isn't the way this is normally done in Adapt.

### DIFF
--- a/js/adapt-stateful-session.js
+++ b/js/adapt-stateful-session.js
@@ -90,7 +90,7 @@ define([
 
     // Session In Progress
     setupEventListeners: function() {
-      $(window).on('beforeunload unload', this._onWindowUnload);
+      $(window).on('beforeunload unload pagehide', this._onWindowUnload);
 
       if (this._shouldStoreResponses) {
         this.listenTo(Adapt.components, 'change:_isInteractionComplete', this.onQuestionComponentComplete);
@@ -109,7 +109,7 @@ define([
     },
 
     removeEventListeners: function () {
-      $(window).off('beforeunload unload', this._onWindowUnload);
+      $(window).off('beforeunload unload pagehide', this._onWindowUnload);
       this.stopListening();
     },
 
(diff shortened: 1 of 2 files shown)


joe-allen-89 (approved):
👍

cahirodoherty-learningpool (approved):
👀

oliverfoster (approved):

oliverfoster:
@canstudios-nicolaw perfect invention of process. I suggest deleting the release branch after doing a manual version release (only the master branch is covered by the automated release process).

Then we can just follow this process any time we need to do similar again, with no leftovers.

PR #317, opened 2024-04-14
Fix: Added cmi.objectives _recordObjectives and id existence check (fixes #316)

### LINKED ISSUE #316: cmi.objectives issues
### Reference

Forum post: https://community.adaptlearning.org/mod/forum/discuss.php?d=4223
LMS: KnowBe4 
Original pr for addition of the cmi.objectives behaviour: https://github.com/adaptlearning/adapt-contrib-spoor/pull/280

> ...KnowBe4 won't let you continue if there are any errors. You can upload your package but cannot add it to any training. The support folks have been good about manually pushing it through for us, but we are going to have monthly training for the next year or so and they always need to be uploaded, viewed by others, then edited and uploaded again. With the need to contact support each time it makes this a lengthy process. That is the impact I was trying to avoid.

### Issues
* No switch
As with cmi.interactions there is a switch to enable/disable the behaviour. There is no such switch for cmi.objectives.

https://github.com/adaptlearning/adapt-contrib-spoor/blob/df90b8dacbe8c20e12e1d57dfd1cdfc37de82903/example.json#L8

* Lazy objective creation
[`getObjectiveById`](https://github.com/adaptlearning/adapt-contrib-spoor/blob/df90b8dacbe8c20e12e1d57dfd1cdfc37de82903/js/scorm/wrapper.js#L696) returns the next objective id if no existing id is found in the objectives list. Subsequently, the id value must be set if a new id is created or the id must be reset unnecessarily if an existing id is found as the code does not differentiate between the two scenarios. This happens after the objective index is fetched at the following lines:

https://github.com/adaptlearning/adapt-contrib-spoor/blob/df90b8dacbe8c20e12e1d57dfd1cdfc37de82903/js/scorm/wrapper.js#L704

https://github.com/adaptlearning/adapt-contrib-spoor/blob/df90b8dacbe8c20e12e1d57dfd1cdfc37de82903/js/scorm/wrapper.js#L713

https://github.com/adaptlearning/adapt-contrib-spoor/blob/df90b8dacbe8c20e12e1d57dfd1cdfc37de82903/js/scorm/wrapper.js#L730

The code should check for objective existence before setting the id so as not to unnecessarily reset it.

oliverfoster:
fixes #316 

### Fix
* Added `_recordObjectives` config
* Check for objective existence before resetting id

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Determines whether the user's responses to questions should be tracked to  the `
 
 ##### \_shouldRecordObjectives (boolean)
 
-Determines whether the user's content objects and their statuses should be tracked to the `cmi.objectives` fields of the SCORM data model or not. Acceptable values are `true` or `false`. The default is `true`. Note that not all SCORM 1.2 conformant Learning Management Systems support `cmi.objectives`. The code will attempt to detect whether support is implemented or not and, if not, will fail gracefully. Occasionally the code is unable to detect when `cmi.objectives` are not supported, in those (rare) instances you can switch off interaction tracking using this property so as to avoid 'not supported' errors.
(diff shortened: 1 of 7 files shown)


cahirodoherty-learningpool (approved):
👀

danielghost (approved):
`cmi.objectives.n.id` shouldn't really need to be checked if it already exists, as it is acceptable to set this to the same value each time. The cost of the additional calls required to loop through the records will be negligable, so worth including if this overcomes LMS implementation issues.

Likewise there are some LMSs that haven't implemented objectives correctly where supported, so good idea to be able to explicitly switch this off for such cases.

joe-allen-89 (approved):
👍

PR #323, opened 2024-12-10
Fix: SCORM 2004 runtime data reset if disconnected on exit

### LINKED ISSUE #320: Course Exit while disconnected from LMS loses progress
Within SCORM 2004, all of my course progress is being erased if I am exiting the course while my internet is disabled. SCORM 1.2 seems to work as you would expect, assuming the data had been committed before the disconnect. SCORM 2004 doesn't and the data is reset unless the cmi.exit state has been set.
 
The other odd thing is that a status doesn't seem to be reflected back into LMS UI until I've relaunched the SCORM. And within certin test LMSs, a potential fix works in Chrome but not Firefox. So a fix would likely need to be tested across a few LMSs & browsers.

### Your environment
* v5.35.3 FW
* SCORM 2004 4th Ed.
* Result varies between browsers
* Windows Laptop

### Steps to reproduce
1. With a SCORM 2004 enabled course, ensure a valid commit has occurred to the LMS.
2. Disable your internet 
3. Allow Adapt's error notification(s) to appear
4. Exit the course
5. Check to see that the LMS UI has been updated appropriately.
6. Enter the course
7. Check to see if progress has been retained.

### Expected behaviour
The LMS UI should be updated with approrpate statuses. Bookmarking should immediately address the issue on relaunch and the appropriate progress should be retained.

danielghost:
Fixes #320.

### Fix
*  Set exit state on initialisation to negate SCORM 2004 runtime data being reset on relaunch if disconnected whilst course window is closed.
* Set session time on each commit to prevent this from potentially being lost if [`finish`](https://github.com/adaptlearning/adapt-contrib-spoor/blob/9dbfaff2ee86abb394304c8471fef80d8e09ee2e/js/scorm/wrapper.js#L315) is not called during exit. This can occur due to being disconnected, or potentially when used via a mobile app which may not trigger the required [window events](https://github.com/adaptlearning/adapt-contrib-spoor/blob/9dbfaff2ee86abb394304c8471fef80d8e09ee2e/js/adapt-stateful-session.js#L129).

### DIFF
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -169,6 +169,7 @@ class ScormWrapper {
 
     this.startTime = new Date();
     this.initTimedCommit();
+    this.setExitState();
     return this.lmsConnected;
   }
 
@@ -289,6 +290,8 @@ class ScormWrapper {
       return;
     }
 
+    this.setSessionTime();
+
     if (this.scorm.save()) {
       this.commitRetries = 0;
       this.lastCommitSuccessTime = new Date();
@@ -335,15 +338,8 @@ class ScormWrapper {
       this.logOutputWin.close();
     }
 
-    this.endTime = new Date();
-
-    if (this.isSCORM2004()) {
-      this.scorm.set('cmi.session_time', this.convertToSCORM2004Time(this.endTime.getTime() - this.startTime.getTime()));
-      this.scorm.set('cmi.exit', this.getExitState());
-    } else {
-      this.scorm.set('cmi.core.session_time', this.convertToSCORM12Time(this.endTime.getTime() - this.startTime.getTime()));
-      this.scorm.set('cmi.core.exit', this.getExitState());
-    }
+    this.setSessionTime();
+    this.setExitState();
 
     if (this._connection) {
       this._connection.stop();
@@ -940,6 +936,15 @@ class ScormWrapper {
     return response.join(',');
   }
 
+  setSessionTime() {
+    const endTime = new Date();
+    if (this.isSCORM2004()) {
+      this.setValue('cmi.session_time', this.convertToSCORM2004Time(endTime.getTime() - this.startTime.getTime()));
+    } else {
+      this.setValue('cmi.core.session_time', this.convertToSCORM12Time(endTime.getTime() - this.startTime.getTime()));
+    }
+  }
+
   getExitState() {
     const completionStatus = this.scorm.data.completionStatus;
     const isIncomplete = completionStatus === COMPLETION_STATE.INCOMPLETE.asLowerCase || completionStatus === COMPLETION_STATE.UNKNOWN.asLowerCase;
@@ -947,6 +952,11 @@ class ScormWrapper {
     if (exitState !== 'auto') return exitState;
     if (this.isSCORM2004()) return (isIncomplete ? 'suspend' : 'normal');
     return '';
+  }
+
+  setExitState() {
+    const dataModel = this.isSCORM2004() ? 'cmi.exit' : 'cmi.core.exit';
+    this.setValue(dataModel, this.getExitState());
   }
 
 }
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -940,9 +940,9 @@ class ScormWrapper {
     const endTime = new Date();
     if (this.isSCORM2004()) {
       this.setValue('cmi.session_time', this.convertToSCORM2004Time(endTime.getTime() - this.startTime.getTime()));
-    } else {
-      this.setValue('cmi.core.session_time', this.convertToSCORM12Time(endTime.getTime() - this.startTime.getTime()));
+      return;
     }
+    this.setValue('cmi.core.session_time', this.convertToSCORM12Time(endTime.getTime() - this.startTime.getTime()));
   }
 
   getExitState() {
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -955,8 +955,8 @@ class ScormWrapper {
   }
 
   setExitState() {
-    const dataModel = this.isSCORM2004() ? 'cmi.exit' : 'cmi.core.exit';
-    this.setValue(dataModel, this.getExitState());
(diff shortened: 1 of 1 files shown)


js/scorm/wrapper.js
```diff
@@ -940,6 +936,15 @@ class ScormWrapper {
     return response.join(',');
   }
 
+  setSessionTime() {
+    const endTime = new Date();
+    if (this.isSCORM2004()) {
+      this.setValue('cmi.session_time', this.convertToSCORM2004Time(endTime.getTime() - this.startTime.getTime()));
+    } else {
+      this.setValue('cmi.core.session_time', this.convertToSCORM12Time(endTime.getTime() - this.startTime.getTime()));
+    }
```
oliverfoster, line 945:
```suggestion
    if (this.isSCORM2004()) {
      this.setValue('cmi.session_time', this.convertToSCORM2004Time(endTime.getTime() - this.startTime.getTime()));
      return;
    }
    this.setValue('cmi.core.session_time', this.convertToSCORM12Time(endTime.getTime() - this.startTime.getTime()));
```

danielghost: I would normally return early to prevent nesting, but it doesn't seem that beneficial in this case due to how the method only contains these conditions. Not bothered either way though if this is preferred in all cases.

oliverfoster: Yes it is preferred for the most part:
https://github.com/adaptlearning/adapt-contrib-spoor/blob/c6fc537b66a72287ac1415609f155ada7db1df67/js/scorm/wrapper.js#L272-L278

Use `if (expression) { } else {` only if there is something after the final `}`, such as:
https://github.com/adaptlearning/adapt-contrib-spoor/blob/c6fc537b66a72287ac1415609f155ada7db1df67/js/scorm/wrapper.js#L648-L657

Otherwise just return early. It's easier to see that the first if block is the end of the function, that scorm2004 is the exception and that the primary use-case is scorm1.2:
```js
    if (this.isSCORM2004()) {
      this.setValue('cmi.session_time', this.convertToSCORM2004Time(endTime.getTime() - this.startTime.getTime()));
      return;
    }
    this.setValue('cmi.core.session_time', this.convertToSCORM12Time(endTime.getTime() - this.startTime.getTime()));
```

`isValidCompletionStatus` is a good example of why return early is easier to read, where one must read 11 extra lines to realise that a `return false` would have been sufficient much earlier:
https://github.com/adaptlearning/adapt-contrib-spoor/blob/c6fc537b66a72287ac1415609f155ada7db1df67/js/scorm/wrapper.js#L749-L772



> Not bothered either way

I've commit the suggestion.

oliverfoster: To make `isValidCompletionStatus` more readable, it could be rewritten like this, where one doesn't have to read ahead to see the extent of the first exception:
```js
isValidCompletionStatus(status) { 
  status = status.toLowerCase(); // workaround for some LMSs (e.g. Arena) not adhering to the all-lowercase rule 
  if (this.isSCORM2004()) { 
    switch (status) { 
      case COMPLETION_STATE.UNKNOWN.asLowerCase: 
      case COMPLETION_STATE.NOTATTEMPTED.asLowerCase: 
      case COMPLETION_STATE.NOT_ATTEMPTED.asLowerCase: // mentioned in SCORM 2004 spec - mapped to 'not attempted' 
      case COMPLETION_STATE.INCOMPLETE.asLowerCase: 
      case COMPLETION_STATE.COMPLETED.asLowerCase: 
        return true; 
    } 
    return false; 
  }
  switch (status) { 
    case COMPLETION_STATE.NOTATTEMPTED.asLowerCase: 
    case COMPLETION_STATE.BROWSED.asLowerCase: 
    case COMPLETION_STATE.INCOMPLETE.asLowerCase: 
    case COMPLETION_STATE.COMPLETED.asLowerCase: 
    case SUCCESS_STATE.PASSED.asLowerCase: 
    case SUCCESS_STATE.FAILED.asLowerCase: 
      return true; 
  }
  return false;
} 
```

or this, which has much fewer keywords for the same output:

```js
isValidCompletionStatus(status) { 
  status = status.toLowerCase(); // workaround for some LMSs (e.g. Arena) not adhering to the all-lowercase rule 
  if (this.isSCORM2004()) { 
    return [
      COMPLETION_STATE.UNKNOWN.asLowerCase,
      COMPLETION_STATE.NOTATTEMPTED.asLowerCase,
      COMPLETION_STATE.NOT_ATTEMPTED.asLowerCase, // mentioned in SCORM 2004 spec - mapped to 'not attempted' 
      COMPLETION_STATE.INCOMPLETE.asLowerCase, 
      COMPLETION_STATE.COMPLETED.asLowerCase
    ].includes(status)
  }
  return [
    COMPLETION_STATE.NOTATTEMPTED.asLowerCase,
    COMPLETION_STATE.BROWSED.asLowerCase,
    COMPLETION_STATE.INCOMPLETE.asLowerCase,
    COMPLETION_STATE.COMPLETED.asLowerCase,
    SUCCESS_STATE.PASSED.asLowerCase, 
    SUCCESS_STATE.FAILED.asLowerCase
  ].includes(status)
}
```

joe-replin (approved):
Looks great! I've tested against the original issue and can confirm that it is working as desired. Thank you for looking into this.

Out of curiosity, should cmi.entry state be listed in the console alongside cmi.exit?

oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

oliverfoster:
Waiting for @danielghost to answer @joe-replin 's question before merging.

lemmyadams (approved):

danielghost:
> Out of curiosity, should cmi.entry state be listed in the console alongside cmi.exit?

It will only log this if [`isFirstSession`](https://github.com/adaptlearning/adapt-contrib-spoor/blob/9dbfaff2ee86abb394304c8471fef80d8e09ee2e/js/scorm/wrapper.js#L178) is called, which isn't the case as default as currently nothing is handled differently depending on the session type. This can be used by other plugins as required.

PR #329, opened 2025-02-25
Fix: remove unwanted `imsss:objectives` from manifest

### LINKED ISSUE #322: SCORM 2004 manifest config with undesired "Results" objective
When analysing `cmi.objectives` for a SCORM 2004 package, the first objective didn't seem to be related to the content. Initially I assumed this was something pre-populate by the LMS, but it seems the imsmanifest we have been using for years (even before SCORM 2004 was a configurable option in Adapt) defines an [objective with the id of "Results"](https://github.com/adaptlearning/adapt-contrib-spoor/blob/9dbfaff2ee86abb394304c8471fef80d8e09ee2e/scorm/2004/imsmanifest.xml#L17-L19).

Would suggest https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/scorm/2004/imsmanifest.xml#L16-L21 and https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/scorm/2004/imsmanifest.xml#L34-L36 should be removed.

danielghost:
Fixes #322.

### Fix
* Remove unwanted `imsss:objectives` from manifest.

### DIFF
--- a/scorm/2004/imsmanifest.xml
+++ b/scorm/2004/imsmanifest.xml
@@ -14,9 +14,6 @@
             <item identifier="I_A001" identifierref="A001">
                 <title><![CDATA[@@course.title]]></title>
                 <imsss:sequencing>
-                    <imsss:objectives>
-                        <imsss:primaryObjective satisfiedByMeasure="false" objectiveID="Results" />
-                    </imsss:objectives>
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

oliverfoster (approved):

PR #332, opened 2025-02-25
Update: Include associated contentObjects with `cmi.interactions.n.objectives`

### LINKED ISSUE #330: Initialise objectives for `_isAvailable` content only
In addition to the changes made in https://github.com/adaptlearning/adapt-contrib-spoor/pull/280, should contentObject objectives be `_isAvailable` aware to account for content that may not be able to be completed?

### LINKED ISSUE #331: Include associated contentObjects to `cmi.interactions.n.objectives`
In addition to https://github.com/adaptlearning/adapt-contrib-spoor/pull/321, adding the associated contentObjects to `cmi.interactions.n.objectives` will provide more context for where the question resides within the course to allow easier mapping of the content for reporting purposes.

danielghost:
Fixes #330.
Fixes #331.

### Update
* Include associated contentObjects for recorded interactions via `cmi.interactions.n.objectives`.
* Amended logic so unavailable contentObjects objectives are recorded with a `completion_status` of 'unknown' for SCORM 2004 and a `status` reuse of 'not attempted' for SCORM 1.2. Any previously recorded scores are maintained unless the associated model has been reset.

### Dependency
 * https://github.com/adaptlearning/adapt_framework/releases/tag/v5.48.2
   * https://github.com/adaptlearning/adapt-contrib-core/pull/670

### DIFF
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -602,7 +602,7 @@ class ScormWrapper {
     } else if (isPercentageBased) {
       // convert values to 0-100 range
       // negative scores are capped to 0 due to SCORM 1.2 limitations
-      score = (score < 0 || (score === 0 && maxScore === 0))
+      score = (score < 0 || maxScore === 0)
         ? 0
         : Math.round((score / maxScore) * 100);
       minScore = 0;


cahirodoherty-learningpool (approved):
👀

oliverfoster:
As discussed on call.

1. Remove dependency on adapt.scoring by having plugins register their applicable objectives onto the questionmodel before recordInteraction. Spoor can then pick up the objectives from the question model 
2. When registering course-wide objectives, extensions should register all objectives initially and use completion_status to indicate their applicability later



Question objectives as follow:
* Add api to questionmodel to add and keep objectives
* Move questionView recordinteraction to after the submitted event https://github.com/adaptlearning/adapt-contrib-core/blob/master/js/views/questionView.js#L214
* Use the submitted event for plugins (adapt-scoring) to register their applicable objectives onto the question model
* Use recordinteraction in spoor and to pass the question model objectives to the scorm 2004 api


Course-wide objective completion_status as follows:

| Status | Description |
| ------ | ------ |
| completed | Required, started and finished |
| incomplete | Required, started |
| not attempted | Required |
| unknown | Not required |


Following scorm 2004 rte:
https://github.com/adaptlearning/scorm_docs/tree/master/SCORM%202004/4th%20Edition
image

danielghost:
This requires further work before going through another round of reviews and ultimately a release.

js/adapt-stateful-session.js
```diff
@@ -179,15 +180,28 @@ export default class StatefulSession extends Backbone.Controller {
...
-      const description = model.get('title') || model.get('displayTitle');
-      offlineStorage.set('objectiveDescription', id, description);
-      if (model.get('_isVisited')) return;
-      const completionStatus = COMPLETION_STATE.NOTATTEMPTED.asLowerCase;
-      offlineStorage.set('objectiveStatus', id, completionStatus);
-    });
+    Adapt.contentObjects.forEach(model => this.initializeContentObjective(model));
+  }
+
+  initializeContentObjective(model) {
```
oliverfoster, line 186: Does this need to be a separate function? Is it called from anywhere else?

js/scorm/wrapper.js
```diff
@@ -595,12 +595,16 @@ class ScormWrapper {
       // range split into negative/positive ranges (rather than minScore-maxScore) depending on score
       const range = (score < 0) ? Math.abs(minScore) : maxScore;
       // `scaled` converted to -1-1 range to indicate negative/positive weighting now that negative values can be assigned to questions
-      const scaledScore = score / range;
+      const scaledScore = (score === 0 && range === 0)
```
oliverfoster, line 598:
```suggestion
      const scaledScore = (range === 0)
```
image

js/adapt-stateful-session.js
```diff
@@ -179,15 +180,28 @@ export default class StatefulSession extends Backbone.Controller {
...
+    if (!this.shouldRecordObjectives || model.get('_recordObjective') === false || model.isTypeGroup('course')) return;
```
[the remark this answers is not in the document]
danielghost: SCORM will already be saving all data against that instance of the SCO, unlike xAPI where it all goes into one store and needs the course to be included to define the context for all statements, not just those using the cmi.interactions profile.
https://github.com/adaptlearning/adapt-contrib-core/blob/e7fe240d9ef552d0f8229bd0ad5f4134d375c43a/js/models/questionModel.js#L421-L425

It can be included, but I don't think it is necessary.

js/adapt-stateful-session.js
danielghost: I think this was left over from an alternative implementation of how _isAvailable content objects were handled:
https://github.com/adaptlearning/adapt-contrib-spoor/commit/fdb9f1550207a866588b05b57d073e6bbed4d668

Happy to combine this now we only ever intialize them once.

oliverfoster: Seems odd having the parent caller be a one liner.

js/scorm/wrapper.js
```diff
@@ -595,12 +595,16 @@ class ScormWrapper {
...
-      const scaledScore = score / range;
+      const scaledScore = (range === 0)
+        ? 0
+        : score / range;
       this.setValue(`${cmiPrefix}.score.scaled`, parseFloat(scaledScore.toFixed(7)));
     } else if (isPercentageBased) {
       // convert values to 0-100 range
       // negative scores are capped to 0 due to SCORM 1.2 limitations
-      score = (score < 0) ? 0 : Math.round((score / maxScore) * 100);
+      score = (score < 0 || (score === 0 && maxScore === 0))
```
oliverfoster, line 605:
```suggestion
      score = (score < 0 || maxScore === 0)
```
Same issue.

js/adapt-stateful-session.js
```diff
@@ -180,16 +181,27 @@ export default class StatefulSession extends Backbone.Controller {
...
     });
   }
 
+  setContentObjectiveStatus(model) {
+    if (!this.shouldRecordObjectives || model.get('_recordObjective') === false || model.isTypeGroup('course')) return;
+    const id = model.get('_id');
+    const isAvailable = model.get('_isAvailable');
+    const isVisited = model.get('_isVisited');
+    const isComplete = model.get('_isComplete');
+    let completionStatus = COMPLETION_STATE.UNKNOWN.asLowerCase;
```
danielghost, line 198: "unknown" is not a valid value for SCORM 1.2. How would we represent not required for unavailable objectives for this spec?

oliverfoster: What would happen if the availability was changed?
Can you delete objectives once they are created?

danielghost: There is no option within SCORM to delete an objective. You could overwite it with another, but not remove it completely.

The options for SCORM 1.2 are likely:
1. Return early if the objective is unavailable. If previously available, the objective would remain in its last recorded state. If not available during initialisation, the objective won't be recorded at all. Probably covers most common use case, but there is an inconsistency between objectives that have always been unavailable and those that become unavailable through the user journey.
2. Use "not attempted" as above. It won't be possible to differentiate between an objective that is mandatory and one that isn't available.
3. Exclude objectives altogether for this version of the spec. It isn't supported by all LMSs for SCORM 1.2. Eliminates any confusion regarding the status but means no objective data, although this was the case until recently anyway in Adapt. Will cause issues for any current courses making use of objectives as it will silently no longer record the data, if the plugin is updated.

Availability changing during runtime is usually more of an edge case, but very possible depending on the design.

Currently any scores already recorded for an objective won't be reset if it becomes unavailable, unless the associated model has also been reset. This is as designed, meaning we don't need to restore a score should it become available again. Those analysing the results can exclude unavailable objectives by looking at the statuses (not possible with option 1).

Is option 2 the easiest to explain to a client? Is the most consistent with our SCORM 2004 implementation.

oliverfoster: I think grouping `not attempted` and `unknown` into one set is the most sensible option. It has reduced accuracy and granularity but it is the best we can muster given the circumstances.

PR #335, opened 2025-03-27
Fix: send session time and exit state during call to finish

### LINKED ISSUE #34: Create properties.schema for origin editor

### LINKED ISSUE #334: cmi.session_time not set before terminate
In the `finish()` method  (wrapper.js), `this.setSessionTime()` is called after `this.finishCalled` is set. 

As a result, the `setValue` call for `cmi.session_time` is skipped, and the final (and most relevant) session time is not recorded. This means the recorded session time will be set to the latest commit time, which is likely out of date or even zero if the course was closed before the first auto-commit.

This can be fixed by calling `this.setSessionTime()` before setting `this.finishCalled` to `true` instead.

danielghost:
Fixes #334

### Fix
* Send session time and exit state during call to finish. Not recorded on exit since https://github.com/adaptlearning/adapt-contrib-spoor/releases/tag/v5.9.10.

### DIFF
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -323,6 +323,8 @@ class ScormWrapper {
       return;
     }
 
+    this.setSessionTime();
+    this.setExitState();
     this.finishCalled = true;
 
     if (this.timedCommitIntervalID !== null) {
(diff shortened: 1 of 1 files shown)


oliverfoster (approved):

joe-allen-89 (approved):
👍

PR #340, opened 2025-05-08
Fix: infinite loop between `setSessionTime` and `commit` when using `commitOnAnyChange`

### LINKED ISSUE #339: `commitOnAnyChange` causes infinite loop
Since https://github.com/adaptlearning/adapt-contrib-spoor/releases/tag/v5.9.10, if `_advancedSettings._commitOnAnyChange: true`, an infinite loop occurs between `setSessionTime()` and `commit()`.

This could be resolved by reverting back to setting the session time via [`scorm.set()`](https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/libraries/SCORM_API_wrapper.js#L567) so it bypasses [`commitOnAnyChange`](https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/js/scorm/wrapper.js#L444) or by amending [`setValue`](https://github.com/adaptlearning/adapt-contrib-spoor/blob/master/js/scorm/wrapper.js#L414-L446) to allow the commit settings to be overriden as required. For example:
```JavaScript
setValue(property, value, shouldCommit = this.commitOnAnyChange) {
  this.logger.debug(`ScormWrapper::setValue: _property=${property} _value=${value}`);
  if (this.finishCalled) {
    this.logger.debug('ScormWrapper::setValue: ignoring request as \'finish\' has been called');
    return;
  }
  if (!this.lmsConnected) {
    this.handleConnectionError();
    return;
  }
  const success = this.scorm.set(property, value);
  if (success) {
    // if success, test the connection as the API usually returns true regardless of the ability to persist the data
    this._connection?.testOnSetValue();
  } else {
    // Some LMSs have an annoying tendency to return false from a set call even when it actually worked fine.
    // So we should only throw an error if there was a valid error code...
    const errorCode = this.scorm.debug.getCode();
    if (errorCode !== 0) {
      this.handleDataError(new ScormError(CLIENT_COULD_NOT_SET_PROPERTY, {
        property,
        value,
        errorCode,
        errorInfo: this.scorm.debug.getInfo(errorCode),
        diagnosticInfo: this.scorm.debug.getDiagnosticInfo(errorCode)
      }));
      return success;
    }
    this.logger.warn('ScormWrapper::setValue: LMS reported that the \'set\' call failed but then said there was no error!');
  }
  if (shouldCommit) this.debouncedCommit();
  return success;
}

setSessionTime() {
  const endTime = new Date();
  if (this.isSCORM2004()) {
    this.setValue('cmi.session_time', this.convertToSCORM2004Time(endTime.getTime() - this.startTime.getTime()), false);
    return;
  }
  this.setValue('cmi.core.session_time', this.convertToSCORM12Time(endTime.getTime() - this.startTime.getTime()), false);
}
```
This would allow us to keep the additional logger calls for help with debugging and diagnostic tests, and additionally handle `commitOnStatusChange` overrides:
```JavaScript
setCompleted() {
  this.setValue(this.isSCORM2004() ? 'cmi.completion_status' : 'cmi.core.lesson_status', COMPLETION_STATE.COMPLETED.asLowerCase, this.commitOnStatusChange);
}
```

danielghost:
Fixes #339.

### Fix
* Infinite loop between `setSessionTime` and `commit` when using `commitOnAnyChange`

### DIFF
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -195,19 +195,19 @@ class ScormWrapper {
     if (this.isSCORM2004()) {
       this.setValue('cmi.completion_status', COMPLETION_STATE.COMPLETED.asLowerCase, options);
       this.setValue('cmi.success_status', SUCCESS_STATE.PASSED.asLowerCase, options);
-    } else {
-      this.setValue('cmi.core.lesson_status', SUCCESS_STATE.PASSED.asLowerCase, options);
+      return;
     }
+    this.setValue('cmi.core.lesson_status', SUCCESS_STATE.PASSED.asLowerCase, options);
   }
 
   setFailed() {
     const options = { shouldCommit: this.commitOnStatusChange || this.commitOnAnyChange };
     if (this.isSCORM2004()) {
-      this.setValue('cmi.success_status', SUCCESS_STATE.FAILED.asLowerCase);
+      this.setValue('cmi.success_status', SUCCESS_STATE.FAILED.asLowerCase, options);
       if (this.setCompletedWhenFailed) this.setValue('cmi.completion_status', COMPLETION_STATE.COMPLETED.asLowerCase, options);
-    } else {
-      this.setValue('cmi.core.lesson_status', SUCCESS_STATE.FAILED.asLowerCase, options);
+      return;
     }
+    this.setValue('cmi.core.lesson_status', SUCCESS_STATE.FAILED.asLowerCase, options);
   }
 
   getStatus() {
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -181,33 +181,33 @@ class ScormWrapper {
   }
 
   setIncomplete() {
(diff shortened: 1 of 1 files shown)


js/scorm/wrapper.js
```diff
@@ -181,33 +181,33 @@ class ScormWrapper {
...
   }
 
   setPassed() {
+    const options = { shouldCommit: this.commitOnStatusChange || this.commitOnAnyChange };
     if (this.isSCORM2004()) {
-      this.setValue('cmi.completion_status', COMPLETION_STATE.COMPLETED.asLowerCase);
-      this.setValue('cmi.success_status', SUCCESS_STATE.PASSED.asLowerCase);
+      this.setValue('cmi.completion_status', COMPLETION_STATE.COMPLETED.asLowerCase, options);
+      this.setValue('cmi.success_status', SUCCESS_STATE.PASSED.asLowerCase, options);
     } else {
```
oliverfoster, line 198: Could you please convert these else blocks to return earlies now that they're not followed by a final line?

oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

PR #348, opened 2025-09-23
Fix: Resetting location via an empty string may be ignored in SCORM 2004

### LINKED ISSUE #344: Resetting location may be ignored in SCORM 2004
There seems to be an issue in the implementation of SCORM 2004 for some LMSs, where setting an empty `cmi.location` is accepted but not restored. Where this is the case, the last populated string will be returned on relaunch, rather than the empty string, resulting in incorrect bookmarking behaviour.

The LMS shouldn't be making any judgement on the format of the string, as its responsibility is to store and return the data determined by the SCO, as long as it is using the correct data type. I am not sure if the affected LMSs are ignoring an empty string as this matches the '403 – Data Model Element Value Not Initialized' used to indicate an unitialized "" location, but this should be a valid string. The LMSs exhibiting this behaviour are returning `true` and a 0 error code on `SetValue()`, indicating the location should be saved correctly.

This issue doesn't appear to be seen for SCORM 1.2, but for consistency it may be worthwhile applying any fix to both.

Whilst this appears to be an LMS implementation issue and not seen on all LMSs, converting empty string values to "null" when being sent to the LMS, and reversed when being retrieved, seems to resolve the issue. This also means the framework and other plugins remain unchanged as they are still expecting the same values as before:

```JavaScript
getLessonLocation() {
  const location = this.getValue(this.isSCORM2004() ? 'cmi.location' : 'cmi.core.lesson_location');
  return location === 'null' ? '' : location;
}

setLessonLocation(location) {
  if (location.trim() === '') location = 'null';
  this.setValue(this.isSCORM2004() ? 'cmi.location' : 'cmi.core.lesson_location', location);
}
```

danielghost:
Fixes #344.

### Fix
* Resetting location via an empty string may be ignored in SCORM 2004 - replaced with "null".

### DIFF
--- a/js/scorm/wrapper.js
+++ b/js/scorm/wrapper.js
@@ -246,10 +246,12 @@ class ScormWrapper {
   }
 
   getLessonLocation() {
-    return this.getValue(this.isSCORM2004() ? 'cmi.location' : 'cmi.core.lesson_location');
+    const location = this.getValue(this.isSCORM2004() ? 'cmi.location' : 'cmi.core.lesson_location');
+    return location === 'null' ? '' : location;
   }
 
   setLessonLocation(location) {
+    if (location.trim() === '') location = 'null';
     this.setValue(this.isSCORM2004() ? 'cmi.location' : 'cmi.core.lesson_location', location);
   }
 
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

ethan-lp (approved):

oliverfoster (approved):

